PR #56, opened 2024-07-22
feat: LWW-Element-Set & Tests

joaopereira12:
LWW-Element-Set & OR-Set implemented in the builtins.


joaopereira12:
I've finished the tests and the implementations. 
I also made the following changes in the existing CRDTs:
- I've fixed the compare method (instead of comparing with '==' it compares the values of each set)
- I've change the ID of the GCounter from string to number (I've read that the ID is a number from 0 to n-1) (https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type#G-Counter_(Grow-only_Counter))

packages/crdt/src/builtins/GCounter/index.ts
```diff
@@ -1,11 +1,19 @@
+export interface IGCounter {
+  globalCounter: number;
+  counts: { [nodeKey: number]: number };
+  value(): number;
+  increment(nodeId: number, amount: number): void;
+  compare(peerCounter: IGCounter): boolean;
+  merge(peerCounter: IGCounter): void;
+}
```
d-roak, line 8: Why?

packages/crdt/src/builtins/GCounter/index.ts
```diff
@@ -1,11 +1,19 @@
...
 /// GCounter with support for state and op changes
 export class GCounter {
   globalCounter: number;
   // instead of standard incremental id for replicas
   // we map the counter with the node id
-  counts: { [nodeId: string]: number };
```
d-roak, line 6: We are mapping nodeId that come from libp2p. They are not numbers

packages/crdt/src/builtins/GCounter/index.ts
```diff
@@ -28,14 +36,19 @@ export class GCounter {
...
+
+  merge(peerCounter: IGCounter): void {
+    let temp: { [nodeKey: number]: number } = Object.assign(
       {},
       this.counts,
       peerCounter.counts,
     );
-    Object.keys(temp).forEach((key) => {
+
+    for (let key in Object.keys(temp)) {
```
d-roak, line 47: Why change?

packages/crdt/src/builtins/GCounter/index.ts
```diff
@@ -28,14 +36,19 @@ export class GCounter {
...
       this.counts[key] = Math.max(this.counts[key], peerCounter.counts[key]);
-    });
+
+      this.globalCounter = Object.values(this.counts).reduce((a, b) => a + b, 0);
```
d-roak, line 50: This shouldn't be inside the loop

packages/crdt/src/builtins/GSet/index.ts
```diff
@@ -19,10 +19,10 @@ export class GSet<T> {
   }
 
   compare(peerSet: GSet<T>): boolean {
-    return this._set === peerSet.set();
+    return (this._set.size == peerSet.set().size && [...this._set].every(value => peerSet.set().has(value)));
```
d-roak, line 22: Doesn't Set type has a native every fn or something that does the same?

d-roak (changes requested):
Initial review

packages/crdt/src/builtins/LWWElementSet/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    lookup(element: T): boolean {
+
+        const addElemTimestmp = this._adds.get(element);
+        const remElemTimestmp = this._removes.get(element);
+
+        if (addElemTimestmp !== undefined &&
+            (remElemTimestmp === undefined || addElemTimestmp > remElemTimestmp)) {
+            return true;
+        }
```
d-roak, line 25: properly write "Timestamp"

packages/crdt/src/builtins/LWWElementSet/index.ts
```diff
@@ -0,0 +1,83 @@
...
+        return false;
+    }
+
+    add(element: T, timestamp: number): void {
```
d-roak, line 29: force timestamp to javascript implementation, don't accept it as an argument (same for other fns)

packages/crdt/src/builtins/LWWElementSet/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    getAdds(): Map<T, number> {
+        return this._adds;
+    }
+
+    getRemoves(): Map<T, number> {
+        return this._removes;
+    }
+
+    areSetsEqual(set1: Set<T>, set2: Set<T>): boolean {
```
d-roak, line 45: put this fn outside the class as a helper fn (private) and rename it to `compareSets`

packages/crdt/src/builtins/LWWElementSet/index.ts
```diff
@@ -0,0 +1,83 @@
...
+        return (this.areSetsEqual(adds, otherAdds) && this.areSetsEqual(rems, otherRems));
+
+    }
+
+    merge(peerSet: LWWElementSet<T>): void {
+
+        for (let element of peerSet._adds.keys()) {
+            if (!this._adds.has(element) || this.compareTimestamp(this._adds.get(element)!, peerSet.getAdds().get(element)!)) {
+                this._adds.set(element, peerSet.getAdds().get(element)!);
+            }
```
d-roak, line 64: why are u making the `!` assertion? also, why not just blindly add the elements?

packages/crdt/src/builtins/LWWElementSet/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    private compareTimestamp(thisTime: number, otherTime: number) {
+
+        const timestampComparison = thisTime - otherTime;
+
+        if (timestampComparison === 0) {
+            return this._bias !== Bias.ADD;
+        }
+        return timestampComparison < 0;
+    }
```
d-roak, line 82: put outside the class

d-roak: and add bias as a parameter

d-roak (changes requested):
Comments on LWWSet

joaopereira12:
This branch also contains some changes in the GSet and contains the GSet and 2PSet tests. I've also added the GCounter and PNCounter to pass the pipeline tests, but I'll add them to their PR with the changes that I made.

PR #89, opened 2024-07-31
feat: Infinite-phase set CDRT and tests

### LINKED ISSUE #80: Implement built-in CRDTs: infinite-phase set
### Background
We already have 2PSet. Infinite-phase set is a generalization of it, allowing elements to be removed and re-added ad infinitum. This should be useful as a built-in CRDT for constructing CROs.

The idea is pretty simple. Use a GCounter to represent the presence of an element. The counter value being odd means the element is present. The counter value being even means the element is not present.

### Task
- implement it
- test it
- document how this type works and its usage

### Resources
- Original paper: https://arxiv.org/abs/2304.01929
- Reference prototype in Go: https://github.com/guiltygyoza/go-crdt/blob/master/infphase_set.go

JanLewDev:
Closes #80


packages/crdt/tests/IPSet.test.ts
```diff
@@ -0,0 +1,144 @@
...
+            set2.remove(peerId, 5);
+            set2.add(peerId, 5);
+
+            set3.add(peerId, 1);
+            set3.remove(peerId, 1);
+            set3.add(peerId, 3);
+
+            console.log(
+                "Merging {1:1, 3:2, 5:3} and {1:2, 3:1} should yield {1:2, 3:2, 5:3}"
+            );
```
d-roak, line 75: rm logging

packages/crdt/tests/IPSet.test.ts
```diff
@@ -0,0 +1,144 @@
...
+        test("Merge Sets Non-Overlapping", () => {
+            set2.add(peerId, 1);
+            set2.add(peerId, 3);
+
+            set3.add(peerId, 5);
+            set3.add(peerId, 7);
+
+            console.log(
+                "Merging {1:1, 3:1} and {5:1, 7:1} should yield {1:1, 3:1, 5:1, 7:1}"
+            );
```
d-roak, line 97: rm logging

d-roak (changes requested):
love it!
change the constructor and it's g2g

d-roak (approved):
lgtm

PR #90, opened 2024-07-31
feat: LWW-Register impl & Tests

### LINKED ISSUE #81: Implement built-in CRDTs: LWW Register
### Background
Register types are useful for CRO to specify a single value that can be updated concurrently by multiple nodes. LWW (Last-Write-Win) Register is a basic type for this purpose.

### Task
- implement
- test

### Resources
- https://lars.hupel.info/topics/crdt/07-deletion/
- https://jsonjoy.com/specs/json-crdt/model-document/crdt-algorithms see the section on "Last-Write-Wins (LWW) CRDT Algorithm"
- https://www.bartoszsypytkowski.com/operation-based-crdts-registers-and-sets/

joaopereira12:
resolves #81

### DIFF
--- a/packages/crdt/src/builtins/LWWRegister/index.ts
+++ b/packages/crdt/src/builtins/LWWRegister/index.ts
@@ -1,4 +1,4 @@
-export type Timestamp = [timestamp: number, replicaId: string];
+type Timestamp = [timestamp: number, replicaId: string];
 
 export class LWWRegister<T> {
     private _element: T;
--- a/packages/crdt/src/builtins/LWWRegister/index.ts
+++ b/packages/crdt/src/builtins/LWWRegister/index.ts
@@ -1,16 +1,16 @@
-export type Timestamp = [timestamp: number, replicaId: number];
+export type Timestamp = [timestamp: number, replicaId: string];
 
 export class LWWRegister<T> {
     private _element: T;
     private _timestamp: Timestamp;
     
 
-    constructor(element: T, replicaId: number) {
+    constructor(element: T, replicaId: string) {
         this._element = element;
         this._timestamp = [Date.now(), replicaId];
     }
 
-    assign(element: T, replicaId: number): void {
+    assign(element: T, replicaId: string): void {
         const timestamp = Date.now();
         if(compareTimestamps(this._timestamp, [timestamp, replicaId])) {
             this._element = element;
--- a/packages/crdt/src/builtins/LWWRegister/index.ts
+++ b/packages/crdt/src/builtins/LWWRegister/index.ts
@@ -1,29 +1,30 @@
-type Timestamp = [timestamp: number, replicaId: string];
-
 export class LWWRegister<T> {
     private _element: T;
-    private _timestamp: Timestamp;
-    
+    private _timestamp: number;
+    private _nodeId: string;
 
-    constructor(element: T, replicaId: string) {
+    constructor(element: T, nodeId: string) {
         this._element = element;
-        this._timestamp = [Date.now(), replicaId];
+        this._timestamp = Date.now();
+        this._nodeId = nodeId;
     }
 
-    assign(element: T, replicaId: string): void {
-        const timestamp = Date.now();
-        if(compareTimestamps(this._timestamp, [timestamp, replicaId])) {
-            this._element = element;
-            this._timestamp = [timestamp, replicaId];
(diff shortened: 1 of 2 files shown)


packages/crdt/src/builtins/LWWRegister/index.ts
```diff
@@ -0,0 +1,43 @@
+type Timestamp = [timestamp: number, replicaId: string];
```
d-roak, line 1: don't use `replicaId`, use `nodeId`

packages/crdt/src/builtins/LWWRegister/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    
+
+    constructor(element: T, replicaId: string) {
+        this._element = element;
+        this._timestamp = [Date.now(), replicaId];
+    }
+
+    assign(element: T, replicaId: string): void {
+        const timestamp = Date.now();
+        if(compareTimestamps(this._timestamp, [timestamp, replicaId])) {
```
d-roak, line 15: the assign is supposed to run locally, the timestamp will never be smaller

packages/crdt/src/builtins/LWWRegister/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    getElement(): T {
+        return this._element;
+    }
+
+    getTimestamp(): Timestamp {
+        return this._timestamp;
+    }
+
+    compare(register: LWWRegister<T>): boolean {
+        return (this._timestamp <= register.getTimestamp());
```
d-roak, line 30: this is wrong

packages/crdt/src/builtins/LWWRegister/index.ts
d-roak: don't create this type. Store both timestamp and nodeId in 2 vars within the class

packages/crdt/src/builtins/LWWRegister/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    merge(register: LWWRegister<T>): void {
+        if(compareTimestamps(this._timestamp, register.getTimestamp())) {
```
d-roak, line 34: extend the if logic here and delete the auxiliary function

d-roak (changes requested):

packages/crdt/src/builtins/LWWRegister/index.ts
joaopereira12: in this paper: https://dsf.berkeley.edu/cs286/papers/crdt-tr2011.pdf - they implement the compare function this way.
But i've seen some other solutions in the resources on the issue that don't have this function.

d-roak: timestamp is not a number, it's a tuple. that's why this is wrong

packages/crdt/tests/LWWRegister.test.ts
```diff
@@ -0,0 +1,57 @@
...
+        expect(register1.getNodeId()).toEqual("node2");
+    });
+
+    // test('Test Compare', () => {
+    //     let register1 = new LWWRegister<string>("alice", "node1");
+    //     let register2 = new LWWRegister<string>("alice", "node2");
+
+    //     expect(register1.compare(register2)).toEqual(true);
+    //     expect(register2.compare(register1)).toEqual(false);
+    // });
```
d-roak, line 21: fix this

packages/crdt/src/builtins/LWWRegister/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    merge(register: LWWRegister<T>): void {
+        const otherTimestamp = register.getTimestamp();
+        const otherNodeId = register.getNodeId();
+        if (otherTimestamp > this._timestamp ||
+            (otherTimestamp === this._timestamp && otherNodeId > this._nodeId)) {
+            this._element = register.getElement();
+            this._timestamp = otherTimestamp;
+            this._nodeId = otherNodeId;
+        }
```
d-roak, line 42: change to return prematurely if:
- the timestamp is bigger
- timestamp equal and nodeId lower

d-roak (changes requested):

d-roak (approved):
lgtm

PR #91, opened 2024-07-31
feat: wasm integration - crdt's and examples refactor

0xMugen:
#6 
# TODO

- [x] Basic assembly script
- [ ] fix the assembly script lol
- [x] GCounter?
- [x] Frontend  
- [ ] Better imports and stucture pipeline from writing to complie and execute in web. 
- [ ] auto compile pipeline? ts => asmscript => wasm

# Overview
This is a replica of the canvas in wasm. objects in the assembly dir get compiled to build and are exposed by the src code, specifically the wasmLoader to hopfuly be used by the front end in a seemless fashion. 

# issues

The objects arn't correctly coded out tbh. currently fixing


0xMugen:
Changed up the crdt-wasm pkgs for it to work under assembly script. No testing is done but wdyt of this design pattern @d-roak ? 
I'm also planning on adding some tests at the pkg level for all the crdts to let us quickly test them. Can't be bothered to build up another from end or manually test it every time.

d-roak:
the asc build will run on the ci, so it will be tested there. We also started having unit tests for the crdt's, they might need to be re-written, but no need to new tests (?)

regarding design pattern, we need some more discussion

d-roak:
closed in favor of #101

PR #99, opened 2024-08-06
feat: LSeq impl & Tests

### LINKED ISSUE #82: Implement built-in CRDTs: LSEQ (array type)
### Background
CRDT has its origin in the collaborative document use case. The document is modeled as an array of characters that can be concurrently edited (character insertion & deletion). There are lots of array/sequence CRDTs for this purpose. The two basic ones are LSEQ and RGA. This issue tracks LSEQ implementation and testing.

### Task
- implement 
- test
- document how this type works and its usage

### Resources
- https://hal.science/hal-00921633/document original paper
- https://www.bartoszsypytkowski.com/operation-based-crdts-arrays-1/
- https://docs.rs/crdts/latest/crdts/list/index.html the Rust crate for CRDT also implements their array type as LSEQ

joaopereira12:
resolves #82

### DIFF
--- a/packages/crdt/src/builtins/LSeq/index.ts
+++ b/packages/crdt/src/builtins/LSeq/index.ts
@@ -1,0 +1,57 @@
+type VPointer = {
+    sequence: Uint8Array,
+    id: string
+};
+
+type Vertex<T> = {
+    vPointer: VPointer,
+    element: T
+}
+
+export class LSeq<T> {
+
+    private _vertices: Array<Vertex<T>>;
+
+    constructor() {
+        this._vertices = new Array<Vertex<T>>();
+    }
+
+    insert(replicaId: string, index: number, content: T): void {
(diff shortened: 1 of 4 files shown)


joaopereira12:
I'm marking this PR as draft because I just realized there's a function I want to change.

packages/crdt/src/builtins/LSeq/index.ts
```diff
@@ -0,0 +1,94 @@
+type VPointer = {
+    sequence: number[],
+    nodeId: string
+};
```
d-roak, line 4:
```suggestion
type VPointer = {
    sequence: number,
    nodeId: string
};
```

joaopereira12: Here the sequence needs to be a byte array. For example, to generate a key between [1] and [2] the result will be [1,1] (https://www.bartoszsypytkowski.com/operation-based-crdts-arrays-1/).

d-roak:
not using crdt builtins anymore

Issue #67, opened 2024-07-29
Improve logging


d-roak:
Currently we are using `console.log` and `console.err` for debugging. We want to create a debug mode where we can check logs, otherwise they shouldn't be displayed.

Research and implement the best approach you can find for this

JanLewDev:
I'm on it.

d-roak:
done at #102 , needs further investigation on how to:
- properly log nested objects without showing too much info
- implement generic logging for types that throw errors

gregemax:
i will like to work on it 

Tell us about your background and how it can be leveraged:
I have a strong background in TypeScript, specializing in creating efficient logging utilities. My experience will help implement a robust solution that enhances maintainability and debugging in our projects.


How would you approach the problem?
Define Debug Mode: Use an environment variable to toggle debug mode.

Create a Logger: Develop a TypeScript logger that wraps console.log and checks debug mode.

Integrate the Logger: Replace existing logging calls for consistency.

Testing: Ensure logs behave as expected in both modes.

d-roak:
Go ahead! Assigned

d-roak:
Any further attempts, please check previous PRs for extra comments. One extra thing that was not mentioned in the last round of comments:
- We want to be able to setup different log levels and configs to each package. The log configs should be passed in the constructor of each package

d-roak:
for tracking #172

gregemax:
hey 
consider a logger like winston or pino with browser-specific handling plugins for more control over object logging and nested data.

d-roak:
hey @gregemax, are you going to continue this task or can i drop the assignment and close the pr?

gregemax:
@d-roak 
should i use  winston or pino  for the logger

d-roak:
winston

d-roak:
assigning it internally

PR #102, opened 2024-08-07
feat: logging improvement

JanLewDev:
#67


packages/network/src/node.ts
```diff
@@ -25,12 +25,33 @@ import { webTransport } from "@libp2p/webtransport";
...
+import { Logger, ILogObj, ISettingsParam } from "tslog";
+
+let mainLogger: Logger<ILogObj> = new Logger();
+
+let startLogger: Logger<ILogObj> = new Logger();
+let subscribeLogger: Logger<ILogObj> = new Logger();
+let unsubscribeLogger: Logger<ILogObj> = new Logger();
+let broadcastMessageLogger: Logger<ILogObj> = new Logger();
+let sendMessageLogger: Logger<ILogObj> = new Logger();
+let sendGroupMessageRandomPeerLogger: Logger<ILogObj> = new Logger();
```
d-roak, line 37: have only, one logger per package. Prepend all the messages from ea pkg with it's name: `topology::{pkg_name}`

packages/network/src/node.ts
```diff
@@ -40,8 +61,11 @@ export class TopologyNetworkNode {
 
   peerId: string = "";
 
-  constructor(config?: TopologyNetworkNodeConfig) {
+  constructor(config: TopologyNetworkNodeConfig) {
     this._config = config;
+    mainLogger = new Logger(config.tslogconfig);
+    mainLogger.settings.name = "topology:network";
```
d-roak, line 67: yep, like this !! do "::" instead

packages/node/src/index.ts
```diff
@@ -10,12 +10,13 @@ import { TopologyObjectStore } from "./store";
...
 import { toString as uint8ArrayToString } from "uint8arrays/to-string";
 import { OPERATIONS } from "./operations.js";
+import tslogconfig from "../tslog.config";
 
 export * from "./operations.js";
 
 // snake_casing to match the JSON config
 export interface TopologyNodeConfig {
-  network_config?: TopologyNetworkNodeConfig;
+  network_config: TopologyNetworkNodeConfig;
```
d-roak, line 19: why?

packages/network/src/node.ts
```diff
@@ -25,12 +25,33 @@ import { webTransport } from "@libp2p/webtransport";
...
+  sendGroupMessageRandomPeerLogger = mainLogger.getSubLogger({ name: "sendGroupMessageRandomTopicPeer" });
+}
+
 // snake_casing to match the JSON config
 export interface TopologyNetworkNodeConfig {
   addresses?: string[];
   bootstrap?: boolean;
   bootstrap_peers?: string[];
   private_key_seed?: string;
+  tslogconfig: ISettingsParam<ILogObj>;
```
d-roak, line 54:
```suggestion
  tslog_config: ISettingsParam<ILogObj>;
```

d-roak: this is setted up with json files, what are the options for it? did you tested with the config files?

tsconfig.json
```diff
@@ -14,6 +14,7 @@
 
     "skipLibCheck": true,
     "stripInternal": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "inlineSourceMap": true
```
d-roak, line 18: why?

packages/node/tslog.config.ts
```diff
@@ -0,0 +1,10 @@
+import { ILogObj, ISettingsParam } from "tslog"
+
+// https://tslog.js.org/#/?id=settings
+const tslogconfig: ISettingsParam<ILogObj> = {
+    minLevel: 3, // 3 = INFO
+    type: "hidden", // "pretty" | "json" | "hidden" 
+    hideLogPositionForProduction: true,
+};
+
+export default tslogconfig;
```
d-roak, line 10: should be in the existing config files

packages/node/tsconfig.json
```diff
@@ -14,5 +14,5 @@
       "path": "../object"
     }
   ],
-  "include": ["src/**/*.ts"]
+  "include": ["src/**/*.ts", "./*.config.ts"]
```
d-roak, line 17: we are dropping the `tslog.config.ts`, you can revert this

packages/node/src/index.ts
```diff
@@ -10,12 +10,13 @@ import { TopologyObjectStore } from "./store";
 import { fromString as uint8ArrayFromString } from "uint8arrays/from-string";
 import { toString as uint8ArrayToString } from "uint8arrays/to-string";
 import { OPERATIONS } from "./operations.js";
+import tslogconfig from "../tslog.config";
 
 export * from "./operations.js";
 
 // snake_casing to match the JSON config
 export interface TopologyNodeConfig {
```
d-roak, line 18: node also has a tslog_config that can be different from the network config

by default, the network will be equal to the node, but they can be set to be different

d-roak (changes requested):

tsconfig.json
JanLewDev: image
so the line is referenced correctly

packages/network/src/node.ts
JanLewDev: should it be a new json file or add to tsconfig.json?

packages/node/tslog.config.ts
JanLewDev: tsconfig.json, right?

JanLewDev: so the tsconfig.json for node would look something like this?
```typescript
{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
    "outDir": "dist"
  },
  "references": [
    {
      "path": "../crdt"
    },
    {
      "path": "../network"
    },
    {
      "path": "../object"
    }
  ],
  "include": ["src/**/*.ts"],
  "tslog_config": {
    "minLevel": 3, // 3 = INFO
    "type": "pretty", // "pretty" | "json" | "hidden" 
    "hideLogPositionForProduction": true
  }
}
```

packages/network/src/node.ts
```diff
@@ -25,12 +25,17 @@ import { webTransport } from "@libp2p/webtransport";
 import { autoNAT } from "@libp2p/autonat";
 import { fromString as uint8ArrayFromString } from "uint8arrays/from-string";
 
+import { Logger, ILogObj, ISettingsParam } from "tslog";
+
+let mainLogger: Logger<ILogObj> = new Logger();
```
d-roak, line 30: rename it to `log`, so we can use `log.info()`

packages/network/src/node.ts
```diff
@@ -25,12 +25,17 @@ import { webTransport } from "@libp2p/webtransport";
 import { autoNAT } from "@libp2p/autonat";
 import { fromString as uint8ArrayFromString } from "uint8arrays/from-string";
```
d-roak, line 27:
```suggestion
```

packages/node/configs/example.json
```diff
@@ -0,0 +1,27 @@
...
+        // Minimum log level to log. 
+        // Options: 0: silly, 1: trace, 2: debug, 3: info, 4: warn, 5: error, 6: fatal
+        "minLevel": 3,
+
+        // Type - hidden || json || pretty
+        "type": "hidden",
+
+        // Hide log position for production
+        "hideLogPositionForProduction": true
+      }
```
d-roak, line 24: only options that exist?

packages/node/tslog.config.ts
d-roak: are you still using this?

packages/node/src/index.ts
```diff
@@ -26,7 +28,19 @@ export class TopologyNode {
 
   constructor(config?: TopologyNodeConfig) {
     this._config = config;
-    this.networkNode = new TopologyNetworkNode(config?.network_config);
+    if(!this._config?.tslog_config){
+      if(!this._config) this._config = {};
```
d-roak, line 32: this shouldn't be inside the prev if

d-roak (changes requested):

d-roak (approved):
lgtm

PR #115, opened 2024-08-20
fix: need to spread the object otherwise you loose the config

sfroment:
We need to spread the object otherwise you'll loose the config assign before

### DIFF
--- a/packages/node/src/index.ts
+++ b/packages/node/src/index.ts
@@ -38,7 +38,7 @@ export class TopologyNode {
       };
     } 
     if(!this._config.network_config?.tslog_config){
-      this._config.network_config = { tslog_config: this._config?.tslog_config };
+      this._config.network_config = { ...this._config.network_config, tslog_config: this._config?.tslog_config };
     }
     this.networkNode = new TopologyNetworkNode(this._config?.network_config);
     this._objectStore = new TopologyObjectStore();


packages/node/src/index.ts
```diff
@@ -38,7 +38,7 @@ export class TopologyNode {
       };
     } 
     if(!this._config.network_config?.tslog_config){
-      this._config.network_config = { tslog_config: this._config?.tslog_config };
+      this._config.network_config = { ...this._config.network_config, tslog_config: this._config?.tslog_config };
```
d-roak, line 41: didn't noticed. what do you think its better?
```
this._config.network_config = { ...this._config.network_config, tslog_config: this._config?.tslog_config };
```
or
```
this._config.network_config.tslog_config = this._config?.tslog_config
```

sfroment: I prefer the latest as we won't recreate the object, but there we can't as at this point in the code we don't know if `this._config.network_config` is not undefined, so we'll have to go for the 1st one if we aren't willing to add 
```js
 if (!this._config.network_config?.tslog_config) {
      if (!this._config.network_config) this._config.network_config = {};
      this._config.network_config.tslog_config = this._config?.tslog_config;
}
```

d-roak: We can go with it, all good

d-roak:
closing due to ts-log revert at #102 / #118

Issue #10, opened 2024-06-16
Fix issue where you have to click twice for connecting to an external CRO


d-roak:
Relevant pieces of code:
https://github.com/topology-foundation/ts-topology/blob/c662c00eb9acabdb50b884191cdfaf7f545a7ad1/examples/canvas/src/index.ts#L71-L87

d-roak:
closed at #116

PR #116, opened 2024-08-20
feat: add object store subscription 

sfroment:
address #10 

https://github.com/user-attachments/assets/92a203dc-6765-41a7-882a-8898415dfd00

This just add a simple way to subscribe to an object id in the topology object store.


d-roak (approved):
love it!

PR #132, opened 2024-08-27
feat: reduce action type for the hashgraph

JanLewDev:


h3lio5 (commented):
Good work on the PR! 
Here's a possible optimization:
```typescript
switch (resolved.action) {
	case ActionType.Reduce:
	         let newOrder = [];
	         for (const idx of resolved.indices) {
		    if (idx === i) shouldIncrementI = false;
	            order[idx] = "";
		 }
		 for (const val of order) {
		     if (val != "") newOrder.push(val);
		 }
		 order = newOrder;
	         if (!shouldIncrementI) j = order.length; // Break out of inner loop
		 break;
	case ActionType.Nop:
		j++;
		break;
}

```

d-roak:
resolve the conflicts after #138 is merged

packages/object/src/hashgraph/index.ts
```diff
@@ -14,8 +16,20 @@ export enum ActionType {
 	DropRight = 1,
 	Nop = 2,
 	Swap = 3,
+	Reduce = 4,
 }
 
+export enum SemanticsType {
+	pairWise = 0,
+	multiVertex = 1,
```
guiltygyoza, line 24: nitpick: change `pairWise` to just `pair`, and change `multiVertex` to just `multiple`

packages/object/tests/hashgraph.test.ts
```diff
@@ -306,3 +301,115 @@ describe("HashGraph for AddWinSet tests", () => {
...
+
+		obj2.merge(obj1.hashGraph.getAllVertices());
+		obj3.merge(obj2.hashGraph.getAllVertices());
+		obj4.merge(obj3.hashGraph.getAllVertices());
+		obj5.merge(obj4.hashGraph.getAllVertices());
+		obj1.merge(obj5.hashGraph.getAllVertices());
+
+		const linearOps = obj1.hashGraph.linearizeOperations();
+		// Pseudo-randomly chosen operation
+		expect(linearOps).toEqual([{ type: "add", value: 3 }]);
```
guiltygyoza, line 413: the 3 here is the one picked by `Smush32` right? would prefer running `Smush32` here and set its result as the expected one, instead of hardcoding 3.

JanLewDev: `Smush32` generates the index based on the seed, which is calculated based on all the concurrent ops

packages/crdt/src/cros/PseudoRandomWinsSet/index.ts
```diff
@@ -0,0 +1,90 @@
...
+	type Hash,
+	type Operation,
+	type ResolveConflictsType,
+	SemanticsType,
+	type Vertex,
+} from "@topology-foundation/object";
+
+const MOD = 1e9 + 9;
+
+function compute_hash(s: string): number {
```
guiltygyoza, line 14: nitpick: snake case vs camel case

packages/crdt/src/cros/PseudoRandomWinsSet/index.ts
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	resolveConflicts(vertices: Vertex[]): ResolveConflictsType {
+		vertices.sort((a, b) => (a.hash < b.hash ? -1 : 1));
+		const seed: string = vertices.map((vertex) => vertex.hash).join("");
+		const rnd = new Smush32(compute_hash(seed));
+		const chosen = rnd.int() % vertices.length;
+		const hashes: Hash[] = vertices.map((vertex) => vertex.hash);
+		hashes.splice(chosen, 1);
+		return { action: ActionType.Reduce, vertices: hashes };
```
guiltygyoza, line 71: I was confused why the chosen one is dropped here. Until I check out how the Reduce action type is handled in `linearizeMultiple`, where the vertices returned by `resolveConflicts` are dropped there and won't show up in the linearization result.

Originally when I thought of reduction as in a lambda function, whose return value stays in the linearization.

If `resolveConflicts` is returning vertices to be dropped, I suggest changing the name of the action type from `Reduce` to `DropMultiple` or simply `Drop`

packages/crdt/src/cros/PseudoRandomWinsSet/index.ts
```diff
@@ -0,0 +1,90 @@
...
+
+	values(): T[] {
+		return Array.from(this.state.entries())
+			.filter(([_, exists]) => exists)
+			.map(([value, _]) => value);
+	}
+
+	resolveConflicts(vertices: Vertex[]): ResolveConflictsType {
+		vertices.sort((a, b) => (a.hash < b.hash ? -1 : 1));
+		const seed: string = vertices.map((vertex) => vertex.hash).join("");
```
guiltygyoza, line 66: nitpick: if the seed only depends on the set of vertices in conflict, and you always initialize the PRNG with this seed and take its first output, the same set of vertices will always result in the same one being chosen. This may not be fair, as in it's "random win" yes but if the same bunch of guys conflict the guy that wins is always the same guy

h3lio5: I guess it is to be deterministic -- seed is always same for the same set of vertices.

guiltygyoza: deterministic yes. I was worried about "same set of ops resolving to the same op winning". I was wrong, the seed is produced from vertex hashes which cover all upstream dependencies. we're good here

h3lio5:
lgtm!

d-roak (approved):
lgtm

guiltygyoza (approved):
lgtm

h3lio5 (approved):

PR #161, opened 2024-09-18
fix: hg deps bug

### LINKED ISSUE #160: Bug with invalid dependencies in the hash graph
There's a bug in the Hash Graph when sending updates in the network related with invalid dependencies. Needs further investigation.

Reproduction:
- Run one example
- Do some actions
- Run a second instance of the example and connect to the same CRO id
- Do some actions with the first instance (it should appear on the second)
- Do some actions with the second instance (it will show errors on the console of the first instance with invalid dependencies)

d-roak:
fixes #160


guiltygyoza (approved):
lgtm.

since `linearizeOperations` doesn't care about the operation carried by root vertex, we can also set rootHash to empty string.

PR #163, opened 2024-09-21
Improve logging

gregemax:
Summary
This PR introduces the debug mode logging functionality for the `Logger` utility. When `DEBUG_MODE` is set to `true`, the `Logger` will output additional debug logs, helping developers track detailed log information during development.

Changes Introduced
- Added a check for `process.env.DEBUG_MODE` in the `Logger.debug` method.
- Implemented two unit tests:
  - One to verify that no logs are printed when `DEBUG_MODE=false`.
  - Another to verify that logs are printed with the correct format when `DEBUG_MODE=true`.

1. Set `DEBUG_MODE=true` in your environment.
2. Use the `Logger.debug()` method to log messages and ensure they appear in the console.
3. Set `DEBUG_MODE=false` and ensure no debug messages are printed.
4. Run `pnpm test` to verify that all unit tests pass.

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -30,7 +30,7 @@
  },
  "devDependencies": {
   "@types/node": "^22.5.4",
-  "tsx": "4.19.0",
+  "tsx": "4.19.1",
   "typescript": "^5.5.4",
   "vitest": "^2.0.5"
  },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -324,8 +324,8 @@ importers:
         specifier: ^22.5.4
         version: 22.5.4
       tsx:
-        specifier: 4.19.0
-        version: 4.19.0
+        specifier: 4.19.1
+        version: 4.19.1
       typescript:
         specifier: ^5.5.4
         version: 5.5.4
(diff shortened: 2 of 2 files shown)


packages/node/logs/debug.log
d-roak, line 1: this file shouldn't be in the repo

d-roak (changes requested):
Hey! Thank you so much for the contribution!

Some considerations:
- The logger should be able to run on the browser (you can't use fs) and on the cli
- It should be configurable with our current config file structure, or when passing the configuration in the node constructor
- We REALLY don't want to "reinvent the wheel" and maintain our own logger if there are alternatives (if we don't have an alternative, we can go with it)
- We want to have different levels of verbosity (ie. info, debug, error, etc etc)
- In the browser, the logs should be nested

Suggestions:
- Look into `ts-log`
- Check #102 
- Check why #102 was reverted (https://github.com/topology-foundation/ts-topology/issues/67#issuecomment-2309080099)
- Check this issue https://github.com/fullstack-build/tslog/issues/295

If you need further clarifications or have any questions, please send us a message on our tg or discord, we answer pretty fast there

PR #164, opened 2024-09-23
refactor: reuse protobuf Vertex definition

### LINKED ISSUE #151: Use `Vertex` from protobuf everywhere
Currently in the HashGraph we have a `Vertex` interface, should be replaced with the protobuf definition

cwkang1998:
Re-use exisiting protobuf Vertex definition for `packages/object`, removing the need to maintain a separate TS typedef. This is done by re-exporting the protobuf-gen typedef with the same name.

Since protobuf makes its message field optional by default since proto3, to address the undefined warning, a few modification is done:

1. `addToFrontier` now returns a Vertex with a defined `operation` field.
2. `merge` now ignores vertex with an undefined operation field. Alternatively we can make `merge` throw or log an error when vertex with undefined operation is received.

Closes #151.

### DIFF
--- a/packages/blueprints/src/AddWinsSet/index.ts
+++ b/packages/blueprints/src/AddWinsSet/index.ts
@@ -44,9 +44,12 @@ export class AddWinsSet<T> implements CRO {
 
 	// in this case is an array of length 2 and there are only two possible operations
 	resolveConflicts(vertices: Vertex[]): ResolveConflictsType {
+		// Both must have operations, if not return no-op
 		if (
-			vertices[0].operation.type !== vertices[1].operation.type &&
-			vertices[0].operation.value === vertices[1].operation.value
+			vertices[0].operation &&
+			vertices[1].operation &&
+			vertices[0].operation?.type !== vertices[1].operation?.type &&
+			vertices[0].operation?.value === vertices[1].operation?.value
 		) {
 			return vertices[0].operation.type === "add"
 				? { action: ActionType.DropRight }
--- a/packages/object/src/hashgraph/index.ts
+++ b/packages/object/src/hashgraph/index.ts
@@ -83,13 +83,13 @@ export class HashGraph {
 	 * @param operation - The operation to be added
 	 * @returns The vertex that was added, with the operation being defined rather than undefined.
 	 */
-	addToFrontier(operation: Operation): Omit<Vertex, "operation"> & { operation: Operation } {
+	addToFrontier(operation: Operation): Vertex {
 		const deps = this.getFrontier();
 		const hash = computeHash(this.nodeId, operation, deps);
 		const vertex = {
 			hash,
 			nodeId: this.nodeId,
-			operation,
+			operation ?? { type: OperationType.NOP },
 			dependencies: deps,
 		};
 
--- a/packages/object/src/hashgraph/index.ts
+++ b/packages/object/src/hashgraph/index.ts
@@ -1,11 +1,13 @@
 import * as crypto from "node:crypto";
 import { linearizeMultiple } from "../linearize/multipleSemantics.js";
 import { linearizePair } from "../linearize/pairSemantics.js";
+import { Vertex_Operation as Operation, Vertex } from "../proto/object_pb.js";
 import { BitSet } from "./bitset.js";
 
+// Reexporting the Vertex and Operation types from the protobuf file
+export { Vertex, Operation };
+
 export type Hash = string;
-// biome-ignore lint: value can't be unknown because of protobuf
-export type Operation = { type: string; value: any | null };
 
 enum OperationType {
 	NOP = "-1",
@@ -29,15 +31,6 @@ export type ResolveConflictsType = {
 	action: ActionType;
 	vertices?: Hash[];
 };
-
-export interface Vertex {
-	hash: Hash;
-	nodeId: string;
-	// internal Operation type enum converted to number
-	// -1 for NOP
-	operation: Operation;
-	dependencies: Hash[];
-}
 
 export class HashGraph {
 	nodeId: string;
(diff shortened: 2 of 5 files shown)


cwkang1998:
Might need to add some tests for the changes I make in `merge`, as its a logical change.

packages/object/src/hashgraph/index.ts
```diff
@@ -86,10 +79,14 @@ export class HashGraph {
 		this.forwardEdges.set(HashGraph.rootHash, []);
 	}
 
-	addToFrontier(operation: Operation): Vertex {
+	/**
+	 * @param operation - The operation to be added
+	 * @returns The vertex that was added, with the operation being defined rather than undefined.
+	 */
+	addToFrontier(operation: Operation): Omit<Vertex, "operation"> & { operation: Operation } {
```
d-roak, line 86: Can safely return `Vertex`

packages/object/src/index.ts
```diff
@@ -103,6 +104,11 @@ export class TopologyObject implements ITopologyObject {
 
 	merge(vertices: Vertex[]) {
 		for (const vertex of vertices) {
+			// Vertex operation is expected to be defined, if its not defined we ignore it and continue.
```
d-roak, line 107:
```suggestion
			// Check to avoid manually crafted `undefined` operations
```

packages/object/src/index.ts
```diff
@@ -112,6 +118,10 @@ export class TopologyObject implements ITopologyObject {
 
 		const operations = this.hashGraph.linearizeOperations();
 		this.vertices = this.hashGraph.getAllVertices().map((vertex) => {
+			// Again, we are filtering out vertices with undefined operations.
```
d-roak, line 121:
```suggestion
```

packages/object/src/index.ts
```diff
@@ -87,6 +87,7 @@ export class TopologyObject implements ITopologyObject {
 
 	// biome-ignore lint: value can't be unknown because of protobuf
 	callFn(fn: string, args: any) {
+		// Since we are creating a vertex with known values, the operation should not be undefined.
```
d-roak, line 90:
```suggestion
```

packages/object/src/hashgraph/index.ts
```diff
@@ -86,10 +79,14 @@ export class HashGraph {
...
 		const deps = this.getFrontier();
 		const hash = computeHash(this.nodeId, operation, deps);
-		const vertex: Vertex = {
+		const vertex = {
 			hash,
 			nodeId: this.nodeId,
 			operation,
```
d-roak, line 92:
```suggestion
			operation ?? { type: OperationType.NOP },
```

d-roak (changes requested):

packages/object/src/hashgraph/index.ts
d-roak, line 86:
```suggestion
	addToFrontier(operation: Operation): Vertex {
```

packages/object/src/index.ts
```diff
@@ -91,10 +91,7 @@ export class TopologyObject implements ITopologyObject {
 		const serializedVertex = ObjectPb.Vertex.create({
 			hash: vertex.hash,
 			nodeId: vertex.nodeId,
-			operation: {
-				type: vertex.operation.type,
-				value: vertex.operation.value,
-			},
+			operation: vertex.operation ?? { type: OperationType.NOP },
```
d-roak, line 94: no need to check type here, we can ensure it won't be null. If there is some edge case where it is, it'll be dropped by the peers and cease to exist

packages/blueprints/src/AddWinsSet/index.ts
```diff
@@ -44,9 +44,12 @@ export class AddWinsSet<T> implements CRO {
 
 	// in this case is an array of length 2 and there are only two possible operations
 	resolveConflicts(vertices: Vertex[]): ResolveConflictsType {
+		// Both must have operations, if not return no-op
 		if (
-			vertices[0].operation.type !== vertices[1].operation.type &&
-			vertices[0].operation.value === vertices[1].operation.value
+			vertices[0].operation &&
+			vertices[1].operation &&
```
d-roak, line 50: why is this needed? you changed correctly the other conditions e.g. `vertices[0].operation?....`

packages/object/src/index.ts
```diff
@@ -111,17 +113,17 @@ export class TopologyObject implements ITopologyObject {
...
-					type: vertex.operation.type,
-					value: vertex.operation.value,
-				},
-				dependencies: vertex.dependencies,
-			};
-		});
+		this.vertices = this.hashGraph
+			.getAllVertices()
+			.filter((vertex) => vertex.operation !== undefined)
+			.map((vertex) => {
```
d-roak, line 119: can you test if we still need this mapping? Since we are using the same type everywhere now, it shouldn't be needed

packages/object/src/index.ts
```diff
@@ -1,8 +1,8 @@
 import * as crypto from "node:crypto";
 import {
-	type ActionType,
 	HashGraph,
 	type Operation,
+	OperationType,
```
d-roak, line 5: import not needed, because of other comment

packages/object/tests/hashgraph.test.ts
d-roak, line 1: love it !!

d-roak (changes requested):

packages/object/src/index.ts
cwkang1998: Yup, we don't need this anymore. Tests seems to pass.

packages/blueprints/src/AddWinsSet/index.ts
d-roak: in theory, it's impossible for that function to receive an undefined operation there (unless u mess with the hg manually). Although i really like your approach and it makes the code more reliable, we can go with it. Although it'll troublesome in the future, to detect that the linearizeOps is returning undefined/null ops

d-roak (approved):
lgtm

PR #172, opened 2024-09-24
Improve logging

gregemax:
flexible logging utility that adapts to both browser and CLI environments, ensuring consistent logging functionality across platforms.


d-roak:
Please read through all the comments and considerations i made in the past. This logger doesn't allow you to log objects at all

If you have questions about the intention behind the feature or need further clarification, please feel free to ask directly in our community group chats or here. I'm more than happy to give you extra clarifications

gregemax:
Hi @d-roak ,

Apologies for the confusion on my last PR. I understand the logger doesn’t fully meet the requirements, especially with object logging.

Could you clarify how the logger should handle object logging (e.g., deep nesting, structured key-value pairs)? Should it behave differently between CLI and browser environments? Are there any preferred libraries or patterns for logging objects? Additionally, could you provide an example of the expected output for logging objects or errors?

If there are any other aspects I should consider or add, please let me know. Thanks for your patience and guidance!

d-roak:
Object expansion only applies to the browser. You can think of it like `console.log()` does. E.g.
image

I don't remember any other consideration apart of all the ones mentioned in the issue and in ur last pr review. Apart of that, the pr is supposed to replace all `console.log` and `console.err` that currently exist in the code base.

Ideally we wanted to go with `ts-log`, but there were some issues with object logging (https://github.com/fullstack-build/tslog/issues/295)

I didn't went through your code this time, i just pulled and tested it. Only after checking that it's working as intended, I'll look into the code

PR #184, opened 2024-10-02
Improving Node Discovery via DHT Announcements - fixes  #152

### LINKED ISSUE #152: Do topic discovery for the pubsub group using a DHT
Currently, we have a big pubsub group where everyone sends messages announcing themselves to discover other peers and the pubsub topics. This doesn't scale.

We want to change it for a DHT that have a `Map<topic_id, nodes[]>`

elielnfinic:
### PR Overview:

This pull request addresses issue #152, focusing on enhancing node discovery for peers subscribed to specific topics by utilizing the DHT (Distributed Hash Table) rather than relying on a centralized PubSub mechanism.

### Summary of Changes:
DHT Integration:  Enabled DHT on the node, allowing topics to map directly to the peers subscribed to them.
Updated `packages/network/node.ts` file:

Added methods: `putDataOnDHT()getDataFromDHT(), announcePeerOnDHT(), removePeerFromDHT(), getPeersOnTopicFromDHT()`

Created a new private property:  `_dht`

Added the following external libraries to the network package:
	•	@libp2p/kad-dht
	•	@libp2p/peer-id
	•	@it-last

### DIFF
--- a/packages/network/package.json
+++ b/packages/network/package.json
@@ -1,63 +1,62 @@
 {
-    "name": "@topology-foundation/network",
-    "version": "0.2.0",
-    "license": "MIT",
-    "repository": {
-      "type": "git",
-      "url": "git+https://github.com/topology-foundation/ts-topology.git"
-    },
-    "type": "module",
-    "types": "./dist/src/index.d.ts",
-    "files": [
-      "src",
-      "dist",
-      "!dist/test",
-      "!**/*.tsbuildinfo"
-    ],
-    "exports": {
-      ".": {
-        "types": "./dist/src/index.d.ts",
-        "import": "./dist/src/index.js"
-      }
-    },
-    "scripts": {
-      "build": "tsc -b",
-      "clean": "rm -rf dist/ node_modules/",
-      "prepack": "tsc -b",
-      "test": "vitest"
-    },
-    "devDependencies": {
-      "@libp2p/interface": "^2.1.0",
-      "react-native-webrtc": "^124.0.3"
-    },
-    "dependencies": {
-      "@bufbuild/protobuf": "^2.0.0",
-      "@chainsafe/libp2p-gossipsub": "^14.1.0",
-      "@chainsafe/libp2p-noise": "^16.0.0",
-      "@chainsafe/libp2p-yamux": "^7.0.0",
-      "@libp2p/autonat": "^2.0.2",
-      "@libp2p/bootstrap": "^11.0.2",
-      "@libp2p/circuit-relay-v2": "^2.0.2",
-      "@libp2p/crypto": "^5.0.2",
-      "@libp2p/dcutr": "^2.0.2",
-      "@libp2p/devtools-metrics": "^1.1.0",
-      "@libp2p/identify": "^3.0.2",
-      "@libp2p/interface-pubsub": "^4.0.1",
-      "@libp2p/kad-dht": "^13.0.2",
-      "@libp2p/mdns": "^11.0.1",
-      "@libp2p/peer-id": "^5.0.2",
-      "@libp2p/peer-id-factory": "^4.2.4",
-      "@libp2p/pubsub-peer-discovery": "^11.0.0",
-      "@libp2p/webrtc": "^5.0.4",
-      "@libp2p/websockets": "^9.0.2",
-      "@libp2p/webtransport": "^5.0.4",
-      "@multiformats/multiaddr": "^12.3.1",
-      "it-length-prefixed": "^9.1.0",
-      "it-map": "^3.1.1",
-      "it-pipe": "^3.0.1",
-      "libp2p": "^2.1.0",
-      "ts-proto": "^2.0.3",
-      "uint8arrays": "^5.1.0"
+  "name": "@topology-foundation/network",
+  "version": "0.2.0",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/topology-foundation/ts-topology.git"
+  },
+  "type": "module",
+  "types": "./dist/src/index.d.ts",
+  "files": [
+    "src",
+    "dist",
+    "!dist/test",
+    "!**/*.tsbuildinfo"
+  ],
+  "exports": {
+    ".": {
+      "types": "./dist/src/index.d.ts",
+      "import": "./dist/src/index.js"
     }
+  },
+  "scripts": {
+    "build": "tsc -b",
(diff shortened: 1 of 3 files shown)


packages/network/package.json
d-roak, line 1: don't change indentation

packages/network/src/node.ts
```diff
@@ -15,9 +15,13 @@ import { generateKeyPairFromSeed } from "@libp2p/crypto/keys";
 import { dcutr } from "@libp2p/dcutr";
 import { devToolsMetrics } from "@libp2p/devtools-metrics";
 import { identify } from "@libp2p/identify";
+
 import type {
 	Ed25519PeerId,
 	EventCallback,
+	EventHandler,
+	PeerId,
+	PeerInfo,
```
d-roak, line 24: not used

packages/network/src/node.ts
```diff
@@ -15,9 +15,13 @@ import { generateKeyPairFromSeed } from "@libp2p/crypto/keys";
 import { dcutr } from "@libp2p/dcutr";
 import { devToolsMetrics } from "@libp2p/devtools-metrics";
 import { identify } from "@libp2p/identify";
+
 import type {
 	Ed25519PeerId,
 	EventCallback,
+	EventHandler,
```
d-roak, line 22: not used

packages/network/src/node.ts
```diff
@@ -26,13 +30,22 @@ import type {
 	StreamHandler,
 	URLPeerId,
 } from "@libp2p/interface";
+
+import {
+	type KadDHT,
+	type ValueEvent,
+	kadDHT,
+	removePublicAddressesMapper,
```
d-roak, line 38: not used

packages/network/src/node.ts
```diff
@@ -26,13 +30,22 @@ import type {
...
+} from "@libp2p/kad-dht";
 import { peerIdFromString } from "@libp2p/peer-id";
 import { pubsubPeerDiscovery } from "@libp2p/pubsub-peer-discovery";
 import { webRTC, webRTCDirect } from "@libp2p/webrtc";
 import { webSockets } from "@libp2p/websockets";
 import { webTransport } from "@libp2p/webtransport";
 import { multiaddr } from "@multiformats/multiaddr";
+import last from "it-last";
 import { type Libp2p, createLibp2p } from "libp2p";
+import { toString as uint8ToString } from "uint8arrays";
```
d-roak, line 48: keep consistency `uint8ArrayToString`

packages/network/src/node.ts
```diff
@@ -97,6 +111,7 @@ export class TopologyNetworkNode {
 				autonat: autoNAT(),
 				dcutr: dcutr(),
 				identify: identify(),
+				dht: kadDHT({}),
```
d-roak, line 114: keep consistency, when not passing configs, don't pass anything as an input

packages/network/src/node.ts
```diff
@@ -248,4 +274,103 @@ export class TopologyNetworkNode {
...
+	}
+
+	/*
+	 * Get the peers on a topic from the DHT
+	 * @param topic The topic to get the peers from
+	 * @returns A set of PeerId
+	 * */
+	async getPeersOnTopicFromDHT(topic: string): Promise<Set<PeerId>> {
+		const uint8Topic = uint8ArrayFromString(topic);
+		const peersOnTopic = await this._dht?.get(uint8Topic);
```
d-roak, line 364: no need for `await`

packages/network/src/node.ts
```diff
@@ -248,4 +274,103 @@ export class TopologyNetworkNode {
...
+			return false;
+		}
+	}
+
+	/**
+	 *
+	 * @param key The key to search for
+	 * @returns The value if the data was found, undefined if the data was not found or the DHT is not initialized
+	 */
+	async getDataFromDHT(key: Uint8Array): Promise<Uint8Array | undefined> {
```
d-roak, line 308: why do we need this method since we have `getPeersOnTopicFromDHT` ?

packages/network/src/node.ts
```diff
@@ -151,6 +167,14 @@ export class TopologyNetworkNode {
 		try {
 			this._pubsub?.subscribe(topic);
 			this._pubsub?.getPeers();
+			this.anouncePeerOnDHT(topic, this._node.peerId);
+
+			// connect to all peers on the topic
+			const peers = await this.getPeersOnTopicFromDHT(topic);
+			for (const peerId of peers) {
+				await this._node.dial(peerId);
```
d-roak, line 175: can't do this. put a `topic_discovery_peers_threshold` variable in the network config and use that threshold as the minimum number of peers.
If it falls down the threshold we try to connect, if it's equal we skip (this is the same everywhere, might segment this logic)

packages/network/src/node.ts
```diff
@@ -248,4 +274,103 @@ export class TopologyNetworkNode {
...
+	 * @param value  The value to search for
+	 * @returns The value `true` if the data was put on the DHT successfully, `false` if not and undefined if the DHT is not initialized
+	 */
+	async putDataOnDHT(
+		key: Uint8Array,
+		value: Uint8Array,
+	): Promise<boolean | undefined> {
+		if (!this._dht) {
+			console.error("DHT not initialized. Please run .start()");
+			return undefined;
```
d-roak, line 290: return false here

packages/network/src/node.ts
```diff
@@ -248,4 +274,103 @@ export class TopologyNetworkNode {
...
+		}
+
+		try {
+			await this._dht?.put(key, value);
+			console.log("Successfully stored data in DHT");
```
d-roak, line 295: change this log to the same format as the other logs

d-roak (changes requested):
partial review

packages/network/src/node.ts
elielnfinic: You are right, I have just tested it.

packages/network/src/node.ts
elielnfinic: Added a test to make sure I can't dial myself.

packages/network/src/node.ts
elielnfinic: I have removed it. Thank you for pointing that out.

packages/network/package.json
d-roak: not resolved

elielnfinic: I have resolved it.

PR #200, opened 2024-10-17
improv: local bootstrap node set up w/ example

elielnfinic:
## Overview 
This PR enhances the Developer Experience (DX) for the local environment. It streamlines the process of running a topology bootstrap on a local device or network, making it easier to connect peers efficiently.

Currently, the changes are applied to the chat example, but if approved, I can extend this implementation to the other examples as well.

## Summary of Changes

Here are the edited files : 
- examples/chat/package.json
- examples/chat/src/index.ts
- packages/network/src/node.ts
- packages/node/configs/node.json
- README.md

Here are the created files : 
- examples/chat/src/vite-env.d.ts
- packages/node/configs/local-bootstrap.json

Feel free to let me know if there are any adjustments or suggestions. Once approved, I can proceed with applying the same improvements to the other examples.


packages/network/src/node.ts
```diff
@@ -63,6 +64,26 @@ export class TopologyNetworkNode {
...
+
+		const peer_discovery = bootstrap_nodes_list.length
+			? [
+					_pubsubPeerDiscovery,
+					bootstrap({
+						list: bootstrap_nodes_list,
+					}),
+				]
+			: [_pubsubPeerDiscovery];
+
```
d-roak, line 86: why this change?

packages/network/src/node.ts
```diff
@@ -102,7 +111,9 @@ export class TopologyNetworkNode {
 				}),
 				webRTC(),
 				webRTCDirect(),
-				webSockets(),
+				webSockets({
+					filter: filters.all,
+				}),
```
d-roak, line 116: why this change?

packages/node/configs/node.json
```diff
@@ -1,9 +1,7 @@
 {
 	"network_config": {
 		"addresses": ["/webrtc"],
-		"bootstrap_peers": [
-			"/ip4/127.0.0.1/tcp/50000/ws/p2p/12D3KooWC6sm9iwmYbeQJCJipKTRghmABNz1wnpJANvSMabvecwJ"
-		],
+		"bootstrap_peers": [],
```
d-roak, line 4: why this change?

examples/chat/src/vite-env.d.ts
d-roak, line 1: this is adding complexity to the example

examples/chat/package.json
```diff
@@ -10,6 +10,7 @@
     "build": "vite build",
     "clean": "rm -rf dist/ node_modules/",
     "dev": "vite serve",
+    "local": "VITE_TOPOLOGY_MODE=local vite serve",
```
d-roak, line 13: adding complexity

README.md
```diff
@@ -52,3 +53,21 @@ pnpm install
...
+- Go to `packages/node` and run the following command to start the node:
+```bash
+pnpm cli relay --config configs/local-bootstrap.json
+```
+Keep the terminal open and copy the `peerId` from the output. You will need it to connect to the node.
+- Open a new terminal and go to `examples/chat` and run the following command to start the example:
+```bash
+pnpm local
+```
+- Make sure you update the `peerId` in the `src/index.ts` file with the one you copied from the node.
```
d-roak, line 72: not clear which index are you talking about here

README.md
```diff
@@ -52,3 +53,21 @@ pnpm install
...
+- You should now be able to see the chat example running in your browser opening the displayed URL.
```
d-roak, line 73: the main readme is not supposed to go through the examples

packages/network/src/node.ts
elielnfinic: If you disconnect WiFi and the `relay.droak.sh` is still in the list of bootstrap_peers, when the relay is run it produces an issue.

packages/network/src/node.ts
elielnfinic: This allows to connect to `ws` if not, you can only connect to `wss`

packages/node/configs/node.json
elielnfinic: This was while testing, I think, I can bring back this since I have `bootstrap-local.json`

examples/chat/package.json
elielnfinic: I thought this would allow to simply run an example with `pnpm local` and connect to the local bootstrap node for [easy node configuration set up](https://github.com/elielnfinic/ts-topology/blob/5fe96b9c8d83ef19d8d144d50e9d2cd682080abd/examples/chat/src/index.ts#L32)

README.md
elielnfinic: My bad, I should add `examples/chat/src/index.ts`

README.md
elielnfinic: Noted, thank you.

packages/network/src/node.ts
d-roak: what do you mean? this change is restructuring code, is not changing logic

packages/network/src/node.ts
d-roak: we are not supposed to allow `ws`, that's right. The question is if it detects that we are working in the localhost, does the ws fail? Localhost usually has special rules (same applies for http/https)

examples/chat/src/vite-env.d.ts
elielnfinic: I have removed this file.

packages/network/src/node.ts
elielnfinic: When there is no bootstrap node list, I shouldn't have the `bootstrap({list : []})` property. 
If it is in the object, I am getting the following error : 
```
@topology-foundation/node@0.2.0 cli /Users/monordi/apps/topology/ts-topology/packages/node
> tsx ./src/run.ts "relay" "--config" "configs/local-bootstrap.json"

/Users/monordi/apps/topology/ts-topology/node_modules/.pnpm/@libp2p+bootstrap@11.0.7/node_modules/@libp2p/bootstrap/src/index.ts:98
      throw new Error('Bootstrap requires a list of peer addresses')
            ^


Error: Bootstrap requires a list of peer addresses```

That's why I created the condition to check if I have bootstrap nodes in my config. If not, pass the `peer-discovery` not having the bootstrap property.

packages/network/src/node.ts
elielnfinic: Yes, `ws` is failing in localhost

d-roak (approved):
lgtm

PR #206, opened 2024-10-28
feat: logging package

winprn:
This package contains a implementation of `BrowserTransport` for browser usage.
Ref: [BrowserConsole](https://github.com/MarcoMedrano/winston-transport-browserconsole/blob/master/src/lib/BrowserConsole.ts)


packages/logger/asconfig.json
d-roak, line 1: can drop this file

packages/logger/package.json
```diff
@@ -0,0 +1,43 @@
...
+  ],
+  "exports": {
+    ".": {
+      "types": "./dist/src/index.d.ts",
+      "import": "./dist/src/index.js"
+    },
+    "./wasm": {
+      "types": "./dist/src/index.d.ts",
+      "import": "./src/index.asc.ts"
+    }
```
d-roak, line 25: can remove it

packages/logger/package.json
```diff
@@ -0,0 +1,43 @@
...
+  },
+  "scripts": {
+    "asbuild": "yarn asbuild:debug && yarn asbuild:release",
+    "asbuild:debug": "asc --config asconfig.json --target debug",
+    "asbuild:release": "asc --config asconfig.json --target release",
```
d-roak, line 30: can remove it

packages/logger/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "build": "tsc -b",
+    "clean": "rm -rf dist/ node_modules/",
+    "prepack": "tsc -b",
+    "test": "vitest"
+  },
+  "devDependencies": {
+    "assemblyscript": "^0.27.29"
+  },
```
d-roak, line 38: can remove it

packages/logger/src/browser.ts
```diff
@@ -0,0 +1,51 @@
...
+			this.emit("logged", entry);
+		});
+
+		const { message, level } = entry;
+		const mappedMethod = this.methods[
+			level as keyof typeof this.methods
+		] as keyof Console;
+
+		if (Object.getOwnPropertySymbols(entry).length >= 2) {
+			// @ts-ignore
```
d-roak, line 31: why ignore the ts compilation stuff?

packages/logger/tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+	"compilerOptions": {
+		"outDir": "dist"
+	},
+	"references": [
+		{
+			"path": "../object"
+		}
+	],
+	"include": ["src/**/*.ts"],
+	"exclude": ["src/**/*.asc.ts"]
```
d-roak, line 12: can rm it

packages/logger/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+	"extends": "../../tsconfig.json",
+	"compilerOptions": {
+		"outDir": "dist"
+	},
+	"references": [
+		{
+			"path": "../object"
+		}
+	],
```
d-roak, line 10: can remove it

packages/logger/src/browser.ts
winprn: I tried some workaround but it seems like this is a problem with TypeScript: https://github.com/Microsoft/TypeScript/issues/24587. The last comment said that the issue was resolved in `4.4.0-dev.20210627`, but somehow it still persists.
If you want to I can do some type conversion like: `entry[Object.getOwnPropertySymbols(entry)[1] as unknown as string];` to remove the ignore.

For the other three, I want to access the key using `[]` notation but it wont let me do that

d-roak: `as keyof winston.LogEntry` should do the trick (or something similar)

winprn:
For the current logging in `package/network`, here're some screenshots about the change:
### Browser
image

### CLI
image

packages/network/src/node.ts
```diff
@@ -52,6 +56,7 @@ export class TopologyNetworkNode {
 
 	constructor(config?: TopologyNetworkNodeConfig) {
 		this._config = config;
+		log = new Logger("topology::network", config?.log_config?.level);
```
d-roak, line 59: pass the `log_config` object instead of just the level

packages/node/src/rpc/index.ts
```diff
@@ -13,6 +14,11 @@ import type {
 } from "../proto/rpc_pb.js";
 
 export function init(node: TopologyNode) {
+	const log = new Logger(
+		"topology::rpc",
+		node.config?.network_config?.log_config,
```
d-roak, line 19: here the same

d-roak (approved):
lgtm

d-roak (approved):
lgtm

Issue #204, opened 2024-10-28
Refactor areCausallyRelatedUsingBFS to Use BFS Correctly and Remove Duplicated Code


trungnotchung:
The `areCausallyRelatedUsingBFS` function incorrectly uses DFS instead of BFS by employing a stack.

Additionally, it duplicates traversal logic when checking for causality in both directions (hash1 -> hash2 and hash2 -> hash1). This function should be refactored to use a queue for BFS and to separate the traversal logic into a reusable helper function to improve readability and maintainability.

`packages/object/src/hashgraph/index.ts`

PR #208, opened 2024-10-28
#204 fix: BFS implementation

trungnotchung:
Refactors the `areCausallyRelatedUsingBFS` method to simplify the BFS implementation for checking causal relationships between vertices.

### DIFF
--- a/packages/object/src/hashgraph/index.ts
+++ b/packages/object/src/hashgraph/index.ts
@@ -255,7 +255,6 @@ export class HashGraph {
 		return false;
 	}
 
-	// Time complexity: O(V), Space complexity: O(V)
 	areCausallyRelatedUsingBFS(hash1: Hash, hash2: Hash): boolean {
 		return (
 			this._areCausallyRelatedUsingBFS(hash1, hash2) ||
@@ -263,7 +262,6 @@ export class HashGraph {
 		);
 	}
 
-	// Time complexity: O(1), Space complexity: O(1)
 	getFrontier(): Hash[] {
 		return Array.from(this.frontier);
 	}
--- a/packages/object/src/hashgraph/index.ts
+++ b/packages/object/src/hashgraph/index.ts
@@ -223,45 +223,44 @@ export class HashGraph {
 		return test1 || test2;
 	}
 
+	private _areCausallyRelatedUsingBFS(start: Hash, target: Hash): boolean {
+		const visited = new Set<Hash>();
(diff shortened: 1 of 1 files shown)


packages/object/src/hashgraph/index.ts
```diff
@@ -223,47 +223,46 @@ export class HashGraph {
...
 
+	// Time complexity: O(V), Space complexity: O(V)
+	areCausallyRelatedUsingBFS(hash1: Hash, hash2: Hash): boolean {
+		return (
+			this._areCausallyRelatedUsingBFS(hash1, hash2) ||
+			this._areCausallyRelatedUsingBFS(hash2, hash1)
+		);
+	}
+
 	// Time complexity: O(1), Space complexity: O(1)
```
d-roak, line 266:
```suggestion
	areCausallyRelated(hash1: Hash, hash2: Hash): boolean {
		return (
			this._areCausallyRelated(hash1, hash2) ||
			this._areCausallyRelated(hash2, hash1)
		);
	}

```

packages/object/src/hashgraph/index.ts
```diff
@@ -223,47 +223,46 @@ export class HashGraph {
 		return test1 || test2;
 	}
 
-	// Time complexity: O(V), Space complexity: O(V)
-	areCausallyRelatedUsingBFS(hash1: Hash, hash2: Hash): boolean {
+	private _areCausallyRelatedUsingBFS(start: Hash, target: Hash): boolean {
```
d-roak, line 226:
```suggestion
	private _areCausallyRelated(start: Hash, target: Hash): boolean {
```

d-roak (changes requested):
rm useless comments and rename the fns
try to build the code, it might need some refactors because the naming would be different

JanLewDev (approved):
Clean logic, I like it. I'm slightly confused with the naming as I think the bitset version should be default and BFS version used when memory consumption can become a problem

packages/object/src/hashgraph/index.ts
d-roak, line 266:
```suggestion
	areCausallyRelatedUsingBFS(hash1: Hash, hash2: Hash): boolean {
		return (
			this._areCausallyRelatedUsingBFS(hash1, hash2) ||
			this._areCausallyRelatedUsingBFS(hash2, hash1)
		);
	}

```

d-roak (approved):
lgtm

PR #212, opened 2024-10-28
Improving Node Discovery via DHT Announcements - fixes #152

### LINKED ISSUE #152: Do topic discovery for the pubsub group using a DHT

elielnfinic:
This PR enhances peer discovery by implementing DHT, addressing issue #152.

The DHT-based discovery mechanism provides greater scalability compared to PubSub and offers additional advantages, including more efficient and persistent peer connections.

In addition to updating the discovery mechanism, the grid example UI has also been updated to reflect this new peer discovery approach.


examples/grid/vite.config.mts
```diff
@@ -1,4 +1,5 @@
 import path from "node:path";
+import { fs } from "memfs";
```
d-roak, line 2: why?

packages/network/src/node.ts
```diff
@@ -132,31 +145,34 @@ export class TopologyNetworkNode {
 
 		if (!this._config?.bootstrap) {
 			for (const addr of this._config?.bootstrap_peers || []) {
-				this._node.dial(multiaddr(addr));
+				const stream = await this._node.dial(multiaddr(addr));
```
d-roak, line 148: when are you using the stream?

packages/network/src/node.ts
```diff
@@ -72,31 +78,36 @@ export class TopologyNetworkNode {
 		const _bootstrapNodesList = this._config?.bootstrap_peers
 			? this._config.bootstrap_peers
 			: [
-					"/dns4/relay.droak.sh/tcp/443/wss/p2p/Qma3GsJmB47xYuyahPZPSadh1avvxfyYQwk8R3UnFrQ6aP",
+					// "/dns4/relay.droak.sh/tcp/443/wss/p2p/Qma3GsJmB47xYuyahPZPSadh1avvxfyYQwk8R3UnFrQ6aP",
+					"/ip4/127.0.0.1/tcp/50000/ws/p2p/12D3KooWC6sm9iwmYbeQJCJipKTRghmABNz1wnpJANvSMabvecwJ",
+					// "/dns4/topology-1.nfinic.com/tcp/4430/wss/p2p/12D3KooWC6sm9iwmYbeQJCJipKTRghmABNz1wnpJANvSMabvecwJ",
```
d-roak, line 83: this can't be merged as is. This assumes that people are running local bootstrap nodes

packages/network/src/node.ts
```diff
@@ -257,4 +273,16 @@ export class TopologyNetworkNode {
...
+	 * This function allows to check if the node is ready to be used on the network
+	 * We check if the node has any multiaddrs as for /webrtc it takes a while to get them assigned from the relay server node.
+	 * @returns boolean
+	 */
+	checkNodeReady(): boolean {
+		if (this._node?.getMultiaddrs().length) {
+			return true;
+		}
+		return false;
+	}
```
d-roak, line 287: not being used

packages/network/src/node.ts
```diff
@@ -132,31 +145,34 @@ export class TopologyNetworkNode {
...
 			"::start: Successfuly started topology network w/ peer_id",
 			this.peerId,
 		);
 
-		this._node.addEventListener("peer:connect", (e) =>
-			log.info("::start::peer::connect", e.detail),
-		);
-		this._node.addEventListener("peer:discovery", (e) => {
+		this._node.addEventListener("peer:connect", async (e) => {
+			console.log("::start::peer::connect", e.detail);
```
d-roak, line 162: this and all `console.log` should be changed to `log`

d-roak (changes requested):
refactor all the examples

examples/grid/vite.config.mts
elielnfinic: Used it at a certain time but forgot to remove it. Will remove it.

packages/network/src/node.ts
elielnfinic: I used it for some tests, will remove it.

packages/network/src/node.ts
elielnfinic: I am using it in the example.

packages/network/src/node.ts
elielnfinic: Sure.

packages/network/src/node.ts
d-roak: ahh, we won't add this function. If the app developers want to check this logic, they should put it in their application. Also the comment is not making sense. The node should have the local multiaddrs as soon as it starts, the relay just adds a few more addresses

packages/network/src/node.ts
```diff
@@ -137,26 +142,26 @@ export class TopologyNetworkNode {
...
 		log.info(
 			"::start: Successfuly started topology network w/ peer_id",
 			this.peerId,
 		);
 
-		this._node.addEventListener("peer:connect", (e) =>
-			log.info("::start::peer::connect", e.detail),
-		);
-		this._node.addEventListener("peer:discovery", (e) => {
+		this._node.addEventListener("peer:connect", async (e) => {});
```
d-roak, line 153: why do we have an empty event listener?

examples/grid/src/index.ts
```diff
@@ -205,11 +199,12 @@ async function main() {
...
-		peers = node.networkNode.getAllPeers();
-		discoveryPeers = node.networkNode.getGroupPeers("topology::discovery");
-		render();
-	});
+	setInterval(() => {
+		if (node.networkNode.checkNodeReady()) {
+			peers = node.networkNode.getAllPeers();
+			render();
+		}
+	}, 1000);
```
d-roak, line 207: having some type of handler is preferable. You are rendering every second even if there are no updates

packages/network/src/node.ts
```diff
@@ -72,31 +72,34 @@ export class TopologyNetworkNode {
 		const _bootstrapNodesList = this._config?.bootstrap_peers
 			? this._config.bootstrap_peers
 			: [
-					"/dns4/relay.droak.sh/tcp/443/wss/p2p/Qma3GsJmB47xYuyahPZPSadh1avvxfyYQwk8R3UnFrQ6aP",
+					"/dns4/topology-1.nfinic.com/tcp/4430/wss/p2p/12D3KooWC6sm9iwmYbeQJCJipKTRghmABNz1wnpJANvSMabvecwJ",
```
d-roak, line 75: Even though you are the founder of https://nfinic.com/, we won't have other companies dependencies in the codebase
Let's setup the bootstrap nodes for topology first and then merge this with those in the hardcoded

d-roak (changes requested):

packages/network/src/node.ts
elielnfinic: Good catch! Sorry, I had to delete that line, I deleted droak.sh line.

d-roak:
close in favor of #251

Issue #213, opened 2024-10-29
Add dotenv support for configuration management


trungnotchung:
Integrate dotenv to load environment variables directly into the config, reducing manual parameter passing and improving flexibility across environments.

PR #215, opened 2024-10-29
#213 integrate dotenv for environment variable management

trungnotchung:
Integrates dotenv to load environment variables directly into the configuration, enhancing flexibility and reducing the need for manual parameter passing across different environments.


elielnfinic:
I don’t think using dotenv is a good approach for loading environment variables in this context, as it is a package that might be included by other JavaScript/TypeScript projects. Once the package is added to a project, it will be located in node_modules, and editing the .env file directly may not be ideal.

d-roak:
> I don’t think using dotenv is a good approach for loading environment variables in this context, as it is a package that might be included by other JavaScript/TypeScript projects. Once the package is added to a project, it will be located in node_modules, and editing the .env file directly may not be ideal.

We had plans to add it. The reasoning you are adding does not make sense, the package is not just to be imported but also to run as a standalone, you want to support the standard ways to configure it

packages/node/src/config.ts
```diff
@@ -0,0 +1,31 @@
+import fs from "node:fs";
+import * as dotenv from "dotenv";
+import { program } from "./cli/index.js";
+
+program.parse(process.argv);
+const opts = program.opts();
```
d-roak, line 6: this is supposed to stay on the `run.ts` we might pass some extra flags that are not config-related

packages/node/src/config.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const initializeEnv = () => {
```
d-roak, line 8: bad naming. Using the config file is not considered using the `env`

packages/node/src/config.ts
```diff
@@ -0,0 +1,31 @@
...
+		network_config: {
+			addresses: process.env.ADDRESSES ? process.env.ADDRESSES.split(",") : [],
+			bootstrap: process.env.BOOTSTRAP === "true",
+			bootstrap_peers: process.env.BOOTSTRAP_PEERS
+				? process.env.BOOTSTRAP_PEERS.split(",")
+				: [],
+			browser_metrics: process.env.BROWSER_METRICS === "true",
+			private_key_seed: process.env.PRIVATE_KEY_SEED,
+		},
+	};
```
d-roak, line 28: the default should be returning `undefined` for using the hardcoded defaults

packages/node/src/config.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export const env = initializeEnv();
```
d-roak, line 31: bad naming. suggestion is to name the const `config`

packages/node/src/config.ts
d-roak: you can pass as an input the `opts.config` as an optional parameter

d-roak: any reasoning to use `const fn = () => {}` instead of `function` ?

trungnotchung: no, i will replace it with `function to` keep the code consistent

packages/node/.env.example
```diff
@@ -0,0 +1,4 @@
+ADDRESSES=/ip4/0.0.0.0/tcp/50000/ws,/ip4/0.0.0.0/tcp/50001
+BOOTSTRAP=true
+BOOTSTRAP_PEERS=
```
winprn, line 3: I think we should have example values for all the fields, instead of leaving them empty

winprn (changes requested):

packages/node/.env.example
trungnotchung: I've updated `.env.example` with example values for all the fields

elielnfinic (commented):
As previously mentionedd, this PR should not be merged because we can't have a .env file in a library that can be included by other projects either in browser or as a node modules.

package.json
```diff
@@ -32,6 +32,7 @@
     "@release-it-plugins/workspaces": "^4.2.0",
     "@types/node": "^22.5.4",
     "assemblyscript": "^0.27.29",
+    "dotenv": "^16.4.5",
```
hoangquocvietuet, line 35: consider remove it

hoangquocvietuet:
> As previously mentionedd, this PR should not be merged because we can't have a .env file in a library that can be included by other projects either in browser or as a node modules.

Dotenv automatically configures settings when we run it locally. However, we can also pass configurations directly into the constructor when creating new topology nodes in other projects. This means topology nodes can operate with either default configurations or custom settings.

package.json
trungnotchung: fixed

elielnfinic (commented):
In standalone environments, looks good to me.

elielnfinic (approved):

packages/node/src/run.ts
```diff
@@ -1,15 +1,12 @@
...
 import { program } from "./cli/index.js";
-import { TopologyNode, type TopologyNodeConfig } from "./index.js";
+import { config, loadConfig } from "./config.js";
+import { TopologyNode } from "./index.js";
 import { init as rpc_init } from "./rpc/index.js";
 
 const run = async () => {
 	program.parse(process.argv);
 	const opts = program.opts();
-	let config: TopologyNodeConfig | undefined;
```
elielnfinic, line 9: Could you please keep config file in this logic?

elielnfinic: Please bring it back so that we have two options `.env` and `config file`.

elielnfinic (changes requested):

packages/node/src/run.ts
trungnotchung: can you explain why? in `loadConfig` i've handled both cases.

elielnfinic: Got it.

elielnfinic (approved):

JanLewDev (approved):
lgtm

winprn:
LGTM

winprn (approved):
LGTM

package.json
JanLewDev: this needs to be resolved to merge

packages/node/.env.example
```diff
@@ -0,0 +1,4 @@
+ADDRESSES=/ip4/0.0.0.0/tcp/50000/ws,/ip4/0.0.0.0/tcp/50001
+BOOTSTRAP=true
+BOOTSTRAP_PEERS=/dns4/relay.droak.sh/tcp/443/wss/p2p/Qma3GsJmB47xYuyahPZPSadh1avvxfyYQwk8R3UnFrQ6aP
+PRIVATE_KEY_SEED=bootstrap
```
d-roak, line 4:
```suggestion
PRIVATE_KEY_SEED=example
```

packages/node/src/config.ts
```diff
@@ -0,0 +1,34 @@
+import fs from "node:fs";
+import * as dotenv from "dotenv";
+import type { TopologyNodeConfig } from "./index.js";
+
+export let config: TopologyNodeConfig = {};
```
d-roak, line 5: don't use global variables, make the loadConfig return the `TopologyNodeConfig` and assign it to a const on the run function in`run.ts`

d-roak (changes requested):

packages/node/.env.example
trungnotchung: updated PRIVATE_KEY_SEED to use 'example' instead of 'bootstrap'.

packages/node/src/config.ts
trungnotchung: changed config to constant instead of a global variable.

packages/node/src/config.ts
```diff
@@ -2,12 +2,14 @@ import fs from "node:fs";
 import * as dotenv from "dotenv";
 import type { TopologyNodeConfig } from "./index.js";
 
-export let config: TopologyNodeConfig = {};
+export function loadConfig(
+	configPath?: string | undefined,
+): TopologyNodeConfig {
+	let config: TopologyNodeConfig = {};
```
d-roak, line 8:
```suggestion
): TopologyNodeConfig | undefined {
	let config: TopologyNodeConfig | undefined;
```

packages/node/src/run.ts
```diff
@@ -1,12 +1,13 @@
...
+import { loadConfig } from "./config.js";
 import { TopologyNode } from "./index.js";
+import type { TopologyNodeConfig } from "./index.js";
 import { init as rpc_init } from "./rpc/index.js";
 
 const run = async () => {
 	program.parse(process.argv);
 	const opts = program.opts();
-	loadConfig(opts.config);
+	const config: TopologyNodeConfig = loadConfig(opts.config);
```
d-roak, line 10:
```suggestion
	const config: TopologyNodeConfig | undefined = loadConfig(opts.config);
```

d-roak (commented):
use `undefined` instead of empty object

d-roak (approved):
lgtm

PR #220, opened 2024-10-29
feat: fix root hash, add DAG testcases, DFS function

hoangquocvietuet:
There is a bug due to an incorrect root hash that causes multiple roots to appear when adding a vertex. Additionally, there are missing test cases for checking DAG compatibility, which I have added. I've also modified the DFS functions, which are invoked during topological sorting.


packages/object/src/hashgraph/index.ts
```diff
@@ -149,26 +150,37 @@ export class HashGraph {
 		return hash;
 	}
 
-	// Time complexity: O(V + E), Space complexity: O(V)
-	topologicalSort(updateBitsets = false): Hash[] {
+	DFS(visited: Map<Hash, number> = new Map()): Hash[] {
```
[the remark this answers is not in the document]
d-roak: why using number instead of `bool`?

packages/object/src/hashgraph/index.ts
```diff
@@ -1,3 +1,4 @@
+import { assert } from "node:console";
```
d-roak, line 1: not used?

packages/object/tests/hashgraph.test.ts
```diff
@@ -1,8 +1,38 @@
...
+	test("Test: HashGraph should be DAG compatibility", () => {
+		const cro1 = obj1.cro as AddWinsSet<number>;
+		const cro2 = obj2.cro as AddWinsSet<number>;
+
+		cro1.add(1);
+		cro2.add(2);
+
+		obj2.merge(obj1.hashGraph.getAllVertices());
+
+		expect(obj2.hashGraph.selfCheckConstraints()).toBe(true);
```
d-roak, line 26: test putting two roots and check if it's false

packages/object/src/hashgraph/index.ts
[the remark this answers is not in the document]
JanLewDev: Visited is set to 1 before calling children and 2 after the loop, not sure of the significance of setting it to 2

packages/object/tests/hashgraph.test.ts
```diff
@@ -1,8 +1,38 @@
+import exp from "node:constants";
+import { before } from "node:test";
```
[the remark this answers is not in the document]
hoangquocvietuet: delete it

packages/object/src/hashgraph/index.ts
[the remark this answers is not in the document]
hoangquocvietuet: If we avoid using numbers and instead set 'visited' to 2, we risk incorrectly detecting a cycle where there isn't one due to the properties of DFS. For example, consider the graph with edges A -> B, A -> C, and B -> C. If we first visit A, then C, and later A -> B -> C, we can't determine if C was visitvisited ined in the current DFS recursion or a previous one. This is why we use the number 2 to update and keep track of the visitation state more accurately.

packages/object/src/hashgraph/index.ts
```diff
@@ -149,26 +150,37 @@ export class HashGraph {
...
-			if (visited.has(hash)) return;
-
-			visited.add(hash);
+			visited.set(hash, 1);
 
 			const children = this.forwardEdges.get(hash) || [];
 			for (const child of children) {
-				visit(child);
+				if (visited.get(child) === 1) {
+					console.error("Cycle detected.");
```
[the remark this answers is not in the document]
hoangquocvietuet: changed

packages/object/src/hashgraph/index.ts
```diff
@@ -149,26 +152,37 @@ export class HashGraph {
...
 		const visit = (hash: Hash) => {
-			if (visited.has(hash)) return;
-
-			visited.add(hash);
+			visited.set(hash, 1);
 
 			const children = this.forwardEdges.get(hash) || [];
 			for (const child of children) {
-				visit(child);
+				if (visited.get(child) === 1) {
```
trungnotchung, line 162: I think these numbers should not be hardcoded. You can try using an enum to prevent misunderstandings.

packages/object/tests/hashgraph.test.ts
```diff
@@ -23,14 +25,43 @@ describe("HashGraph construction tests", () => {
 
 		obj2.merge(obj1.hashGraph.getAllVertices());
 
-		expect(obj2.hashGraph.selfCheckConstraints()).toBe(true);
+		// expect(obj2.hashGraph.selfCheckConstraints()).toBe(true);
```
winprn, line 28: consider delete the line?

hoangquocvietuet: readd it

packages/object/src/hashgraph/index.ts
hoangquocvietuet: add it

packages/object/src/hashgraph/index.ts
```diff
@@ -7,11 +8,19 @@ import {
...
 import { BitSet } from "./bitset.js";
 
+const log: Logger = new Logger("hashgraph");
+
 // Reexporting the Vertex and Operation types from the protobuf file
 export { Vertex, Operation };
 
 export type Hash = string;
 
+export enum DepthFirstSearchState {
```
trungnotchung, line 18: enum values should be uppercas

hoangquocvietuet: done

packages/object/src/hashgraph/index.ts
```diff
@@ -149,26 +158,44 @@ export class HashGraph {
...
+			visited.set(hash, DepthFirstSearchState.VISITED);
 		};
-		// Start with the root vertex
+
 		visit(HashGraph.rootHash);
+
+		return result;
+	}
+
+	// Time complexity: O(V + E), Space complexity: O(V)
```
trungnotchung, line 193: you should remove the useless comments, as Oak suggested

packages/object/src/hashgraph/index.ts
```diff
@@ -118,7 +127,7 @@ export class HashGraph {
 			return hash; // Vertex already exists
 		}
 
-		// Temporary fix: don't add the vertex if the dependencies are not present in the local HG.
+		// Temporary fix: don't add the vertex if the dependencies are not present in the local HG.dw
```
trungnotchung, line 130: i guess this is a typo mistake

hoangquocvietuet: removed

packages/object/src/hashgraph/index.ts
hoangquocvietuet: i think i would be fine for later contributor

d-roak: yes, remove it pls

trungnotchung:
lgtm

trungnotchung (approved):
lgtm

packages/object/src/hashgraph/index.ts
hoangquocvietuet: removed

winprn (changes requested):

packages/object/src/hashgraph/index.ts
```diff
@@ -154,22 +160,29 @@ export class HashGraph {
...
+		}
 		const visit = (hash: Hash) => {
-			visited.set(hash, 1);
+			visited.set(hash, DepthFirstSearchState.Visiting);
 
 			const children = this.forwardEdges.get(hash) || [];
 			for (const child of children) {
-				if (visited.get(child) === 1) {
+				if (visited.get(child) === DepthFirstSearchState.Visiting) {
 					log.error("::hashgraph::DFS: Cycle detected");
```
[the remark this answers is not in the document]
d-roak: ?? isn't "::DFS"?

winprn (approved):
Let's ship it! 🚀

JanLewDev (approved):
lgtm

PR #226, opened 2024-11-01
feat: optimize linearization functions

hoangquocvietuet:
The issue mentioned that the linearization was messy and performed poorly, so I created a new PR to optimize it. During this process, I also found a bug in the optimization progress, specifically when performing the swap operation. The bug occurs because the bitset is not swapped. I’ve fixed that as well.

### DIFF
--- a/packages/object/src/hashgraph/bitset.ts
+++ b/packages/object/src/hashgraph/bitset.ts
@@ -99,4 +99,10 @@ export class BitSet {
 		}
 		return undefined;
 	}
+
+	swap(i: number, j: number): void {
+		const temp = this.get(i);
+		this.set(i, this.get(j));
+		this.set(j, temp);
+	}
 }
--- a/packages/object/src/hashgraph/bitset.ts
+++ b/packages/object/src/hashgraph/bitset.ts
@@ -82,9 +82,9 @@ export class BitSet {
 			.join("");
 	}
 
-	findNext(index: number, bit: number): number {
-		let byteIndex = (index / 32) | 0;
-		let bitIndex = index % 32;
+	findNext(index: number, bit: number): number | undefined {
+		let byteIndex = ((index + 1) / 32) | 0;
+		let bitIndex = (index + 1) % 32;
 		let mask = 1 << bitIndex;
 		while (byteIndex < this.data.length) {
 			while (bitIndex < 32) {
@@ -97,6 +97,6 @@ export class BitSet {
 			bitIndex = 0;
 			mask = 1;
 		}
-		return -1;
+		return undefined;
 	}
 }
--- a/packages/object/src/hashgraph/bitset.ts
+++ b/packages/object/src/hashgraph/bitset.ts
@@ -81,4 +81,22 @@ export class BitSet {
 			.map((int) => int.toString(2).padStart(32, "0"))
 			.join("");
 	}
+
+	findNext(index: number, bit: number): number {
+		let byteIndex = (index / 32) | 0;
+		let bitIndex = index % 32;
+		let mask = 1 << bitIndex;
+		while (byteIndex < this.data.length) {
+			while (bitIndex < 32) {
+				if ((this.data[byteIndex] & mask) === bit)
+					return byteIndex * 32 + bitIndex;
+				mask <<= 1;
+				bitIndex++;
+			}
+			byteIndex++;
+			bitIndex = 0;
+			mask = 1;
+		}
+		return -1;
+	}
 }
--- a/packages/object/src/hashgraph/bitset.ts
+++ b/packages/object/src/hashgraph/bitset.ts
@@ -89,9 +89,7 @@ export class BitSet {
 
 		while (wordIndex < this.data.length) {
 			let currentWord = this.data[wordIndex];
-
 			if (bit === 0) currentWord = ~currentWord;
-
 			currentWord &= mask;
 
 			if (currentWord !== 0) {
--- a/packages/object/src/hashgraph/bitset.ts
+++ b/packages/object/src/hashgraph/bitset.ts
@@ -6,103 +6,106 @@
 	Algorithm for more optimal causality check inspired by https://stackoverflow.com/a/78133041
 */
 export class BitSet {
-	private data: Uint32Array;
+  private data: Uint32Array;
 
-	constructor(size = 1) {
-		// Always start with size 32
-		this.data = new Uint32Array(size);
-	}
+  constructor(size = 1) {
+    // Always start with size 32
+    this.data = new Uint32Array(size);
+  }
 
-	clear(): void {
-		this.data = new Uint32Array(this.data.length);
-	}
+  clear(): void {
+    this.data = new Uint32Array(this.data.length);
+  }
 
-	set(index: number, value: boolean): void {
-		// (index / 32) | 0 is equivalent to Math.floor(index / 32)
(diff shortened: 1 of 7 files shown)


packages/object/tests/hashgraph.test.ts
winprn, line 1: why the comments got formatted?

packages/object/src/linearize/pairSemantics.ts
```diff
@@ -29,31 +40,32 @@ export function linearizePair(hashGraph: HashGraph): Operation[] {
...
+						break;
 					case ActionType.DropRight:
-						order.splice(j, 1);
-						continue; // Continue with the same j
+						dropped[j] = true;
+						break;
 					case ActionType.Swap:
 						[order[i], order[j]] = [order[j], order[i]];
-						j = order.length; // Break out of inner loop
+						j = i;
```
winprn, line 50: why would it be `j = i` instead of the original `j = order.length`

packages/object/src/hashgraph/index.ts
```diff
@@ -258,6 +258,19 @@ export class HashGraph {
 		return false;
 	}
 
+	findNextUnusuallyRelated(hash: Hash, start: number): number | undefined {
+		if (!this.arePredecessorsFresh) {
+			this.topologicalSort(true);
+		}
+		const currentIndex = this.topoSortedIndex.get(hash);
+		if (currentIndex === undefined) return undefined;
```
winprn, line 266: I think you can just check the logic in one statement, no need for extra variable

winprn (changes requested):

packages/object/tests/hashgraph.test.ts
hoangquocvietuet: idk the biome toolchain did

packages/object/src/hashgraph/index.ts
```diff
@@ -258,6 +258,19 @@ export class HashGraph {
 		return false;
 	}
 
+	findNextUnusuallyRelated(hash: Hash, start: number): number | undefined {
```
JanLewDev, line 261:
```suggestion
	findNextCausallyUnrelated(hash: Hash, start: number): number | undefined {
```

JanLewDev (changes requested):
This approach can also speed up `linearizeMultiple`

packages/object/src/linearize/pairSemantics.ts
hoangquocvietuet: To break the inner loop, we set j = order.length. However, since this is a swap operation between i and j, nothing is dropped and we can continue with the same i. For the loop to proceed, we need to find the next causally unrelated index for j, thus we set j = findNextCausallyUnrelated(i) to continue the loop.

packages/object/src/hashgraph/index.ts
hoangquocvietuet: changed

packages/object/src/hashgraph/index.ts
hoangquocvietuet: changed

packages/object/src/linearize/pairSemantics.ts
```diff
@@ -6,15 +6,26 @@ import {
...
+			i++;
+			continue;
+		}
 		const anchor = order[i];
-		let j = i + 1;
-		let shouldIncrementI = true;
+		let j: number = i + 1;
 
 		while (j < order.length) {
+			if (dropped[i]) break;
```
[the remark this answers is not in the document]
hoangquocvietuet: Yes, exactly. Imagine the case when ActionType.DropLeft occurs.

packages/object/tests/hashgraph.test.ts
winprn: please help reverting it 🙏

packages/object/src/linearize/pairSemantics.ts
trungnotchung: i think we shouldn’t continue with j = i. In the previous approach, we broke the inner loop but continued with the same i to ensure we iterate over all vertices in topological sort order.

packages/object/src/linearize/pairSemantics.ts
```diff
@@ -29,31 +40,32 @@ export function linearizePair(hashGraph: HashGraph): Operation[] {
...
+						break;
 					case ActionType.DropRight:
-						order.splice(j, 1);
-						continue; // Continue with the same j
+						dropped[j] = true;
+						break;
 					case ActionType.Swap:
-						[order[i], order[j]] = [order[j], order[i]];
-						j = order.length; // Break out of inner loop
+						reachableState.swap(i, j);
```
trungnotchung, line 49: it doesn’t solve the problem when we want to swap the positions of order[i] and order[j] also requires updating the reachablePredecessors of other nodes. maybe this approach does not work with current pairSemantics.

packages/object/src/hashgraph/index.ts
```diff
@@ -187,39 +253,11 @@ export class HashGraph {
...
-						result[i],
-						reachable?.or(depReachable) || depReachable,
-					);
-				}
-			}
-		}
-
-		this.arePredecessorsFresh = true;
-		return result;
+		return new ReachableState(result, this.vertices);
```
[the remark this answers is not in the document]
JanLewDev: I agree with this. Imo it overcomplicates things and doesn't actually make `ActionType.Swap` work

packages/object/src/linearize/pairSemantics.ts
hoangquocvietuet: Setting j = i allows us to continue with the same i as in the previous approach and effectively resets j. It’s the same process.

packages/object/src/linearize/pairSemantics.ts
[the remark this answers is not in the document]
hoangquocvietuet: The break keyword only exits the switch case, not the outer loop. Please make a note of this.

packages/object/src/hashgraph/bitset.ts
```diff
@@ -81,4 +81,28 @@ export class BitSet {
...
+			while (bitIndex < 32) {
+				if ((this.data[byteIndex] & mask) === bit)
+					return byteIndex * 32 + bitIndex;
+				mask <<= 1;
+				bitIndex++;
+			}
+			byteIndex++;
+			bitIndex = 0;
+			mask = 1;
+		}
```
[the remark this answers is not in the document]
hoangquocvietuet: How can we find the first zero bit in O(1) time? Using bitwise operations still results in O(n/32) complexity due to looping through byteIndex.

packages/object/src/hashgraph/index.ts
```diff
@@ -316,6 +315,10 @@ export class HashGraph {
 		return true;
 	}
 
+	findNextUnusuallyRelated(hash: Hash, start: number): number | undefined {
```
JanLewDev, line 318:
```suggestion
	findNextCausallyUnrelated(hash: Hash, start: number): number | undefined {
```
it was already changed, but reverted?

trungnotchung: i changed the function name

hoangquocvietuet:
The logic part looks good to me.

JanLewDev (approved):
lgtm

d-roak (approved):

PR #241, opened 2024-11-07
feat: add grid collision example

### LINKED ISSUE #228: Modify one of the examples to showcase resolveConflicts()
Currently none of the examples (canvas, chat, grid) demonstrates the use of resolveConflicts(). One important purpose of examples is to showcase the programmability of CROs, where conflict resolution strategy plays a key role in enabling richer CRO behavior.

We already have three examples, whose CROs are designed in similar ways. Suggestion here is to modify one of the examples to showcase meaningful use of conflict resolution.

cwkang1998:
Add a new example with grid 2d where collision is implemented. This utilizes the `resolveConflicts` CRO function for conflict resolution on the user position.

This also reveals an issue with the `this` context within the `object` package when `resolveConflicts` is callback from `TopologyObject`. Opened an issue at #242.

Closes #228


cwkang1998:
Facing some issues running this and getting it working with local bootstrap node. Will mark as done once tested to be working on local.

d-roak:
closing as not been tackled rn

Issue #242, opened 2024-11-07
Incorrect `this` context for `resolveConflict` causing CRO to lose context to itself


cwkang1998:
During runtime, when attempting to access CRO methods within the scope of `resolveConflict`, `this` loses its context and no longer refers to the implemented CRO but rather points to the hashgraph (i.e. `this` === `HashGraph`). This is caused by passing `cro.resolveConflicts` during the hash graph construction in the `TopologyObject`.

Fixing this is crucial to allow implementors of the CRO interface to access their implementation methods and states.

The problematic line of code is located here: https://github.com/topology-foundation/ts-topology/blob/cda5dd5300be6656f0b53738fa9e1c1054835d80/packages/object/src/index.ts#L59

d-roak:
fixed with #243

PR #243, opened 2024-11-07
fix: lost of CRO context in `resolveConflicts`

cwkang1998:
Per title, fix issue #242


elielnfinic (approved):
I agree that context can be lost at certain time. Good to pass the cro in bind.

d-roak (approved):
nice catch

PR #246, opened 2024-11-11
feat: add new bootstrap nodes; sort order for dialing discovered ma's

d-roak:
props to @sfroment for the original sorting function for multiaddrs

### DIFF
--- a/packages/network/src/node.ts
+++ b/packages/network/src/node.ts
@@ -72,7 +72,8 @@ export class TopologyNetworkNode {
 		const _bootstrapNodesList = this._config?.bootstrap_peers
 			? this._config.bootstrap_peers
 			: [
-					"/dns4/relay.droak.sh/tcp/443/wss/p2p/Qma3GsJmB47xYuyahPZPSadh1avvxfyYQwk8R3UnFrQ6aP",
+					"/dns4/bootstrap1.topology.gg/tcp/443/wss/p2p/12D3KooWBu1pZ3v2u6tXSmkN35kiMLENpv3bEXcyT1GJTVhipAkG",
+					"/dns4/bootstrap2.topology.gg/tcp/443/wss/p2p/12D3KooWLGuTtCHLpd1SBHeyvzT3kHVe2dw8P7UdoXsfQHu8qvkf",
 				];
 
 		const _pubsubPeerDiscovery = pubsubPeerDiscovery({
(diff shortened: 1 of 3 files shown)


packages/network/src/node.ts
```diff
@@ -149,9 +150,25 @@ export class TopologyNetworkNode {
...
+				const localRegex =
+					/(^127\.)|(^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.)/;
+				const aLocal = localRegex.test(a.toString());
+				const bLocal = localRegex.test(b.toString());
+				const aWebrtc = a.toString().includes("/webrtc/");
+				const bWebrtc = b.toString().includes("/webrtc/");
+				if (aLocal && !bLocal) return 1;
+				if (!aLocal && bLocal) return -1;
+				if (aWebrtc && !bWebrtc) return -1;
+				if (!aWebrtc && bWebrtc) return 1;
```
trungnotchung, line 163: to ensure WebRTC addresses are dialed before non-local ones, i think we should update the condition order

packages/network/src/node.ts
```diff
@@ -149,9 +150,25 @@ export class TopologyNetworkNode {
...
+				return 0;
+			});
+
+			// Dial WebRTC addresses first, then non-local addresses
+			for (const ma of sortedAddrs) {
```
trungnotchung, line 168: consider renaming `ma` to `address` for clarity

packages/network/src/node.ts
d-roak: ordering imo should be:
- all non-local
  - all webrtc
  - all non-webrtc
- all local
  - all webrtc
  - all non-webrtc

wdyt?
also, is the logic compliant with that ordering?

packages/network/src/node.ts
d-roak: do you know how to create suggestions while reviewing? for this types of comment u can do straight a suggestion
image

packages/network/src/node.ts
trungnotchung: `Dial WebRTC addresses first, then non-local addresses`: this comment in line 167 makes me think that the order looks like:
* all webrtc 
 * all non-local
 * all local
* all non-webrtc
 * all non-local
 * all local

d-roak: let's change the comment then

elielnfinic: When testing, this is the order I am getting : 
- webrtc
   - local
   - non-local 
- non-webrtc
  - local
  - non-local

E.g 
Input :  ```[
    "/ip4/127.0.0.1/tcp/4001/p2p/12D3KooWQJ",
    "/ip4/127.0.0.1/tcp/4001/webrtc/p2p/12D3KooWQJ",
    "/ip4/192.168.0.1/tcp/4001/webrtc/p2p/12D3KooWQJ",
    "/ip4/172.20.34.1/tcp/4001/webrtc/p2p/12D3KooWQJ",
    "/ip4/20.34.1/tcp/4001",
    "/ip4/20.34.1/tcp/4001/webrtc/p2p/12D3KooWQJ",
]```

Output : ```[
  '/ip4/127.0.0.1/tcp/4001/webrtc/p2p/12D3KooWQJ',
  '/ip4/192.168.0.1/tcp/4001/webrtc/p2p/12D3KooWQJ',
  '/ip4/172.20.34.1/tcp/4001/webrtc/p2p/12D3KooWQJ',
  '/ip4/20.34.1/tcp/4001/webrtc/p2p/12D3KooWQJ',
  '/ip4/127.0.0.1/tcp/4001/p2p/12D3KooWQJ',
  '/ip4/20.34.1/tcp/4001'
]```

elielnfinic (changes requested):

packages/network/src/node.ts
d-roak: good catch, the regex is wrong

elielnfinic (approved):
Sort is working perfectly.

trungnotchung (approved):

Issue #185, opened 2024-10-02
Bad dependencies if peers discover each other after both connecting to the same CRO


guiltygyoza:
Bad dependencies cause the vertices sent by one peer to be rejected by another peer. 

https://github.com/user-attachments/assets/ae3bd8e7-d69e-4af9-8ad6-3acc51e96a35

d-roak:
not fixed
<img width="1494" alt="image" src="https://github.com/user-attachments/assets/3a1b74d3-0fde-4c2b-aec8-71ebad1418a0">

elielnfinic:
noted

d-roak:
fixed

PR #247, opened 2024-11-12
Fixing vertex dependencies issue

elielnfinic:
This PR fixes issue #185 
When a peer joins a CRO, it synchronizes the two peers vertex before merging them.


elielnfinic (commented):
I know some code are weird on try and catches, please let me know what you think.

packages/node/src/handlers.ts
```diff
@@ -173,7 +185,21 @@ function syncAcceptHandler(
...
+				object.merge(vertices);
+			} catch (err) {
+				console.error(
+					"topology::node::syncAcceptHandler",
+					"Error merging vertices",
+					err,
+				);
+				return;
+			}
+		}
```
d-roak, line 202: Was there a bug here? For sure you are not supposed to restart the sync process in the middle of it

d-roak, line 202: this code is crazy
if() -> catch() -> try()

not clean at all

packages/node/src/handlers.ts
```diff
@@ -78,20 +78,32 @@ function updateHandler(node: TopologyNode, data: Uint8Array) {
...
+					dependencies: v.dependencies,
+				};
+			}),
+		);
+	};
+
+	try {
+		_merge();
+	} catch (err) {
+		if (updateMessage.vertices.length === 0) return false;
```
d-roak, line 100: why is this check here? it's independent of the merge operation

packages/node/src/handlers.ts
```diff
@@ -78,20 +78,32 @@ function updateHandler(node: TopologyNode, data: Uint8Array) {
...
+
+		const peerId = updateMessage.vertices[0].nodeId;
+		node.syncObject(object.id, peerId);
+		_merge();
```
d-roak, line 104: syncObject is async, it will be sent over the network. Doesn't make sense to have the merge being called after

packages/object/src/hashgraph/index.ts
```diff
@@ -128,8 +128,7 @@ export class HashGraph {
 		if (
 			!deps.every((dep) => this.forwardEdges.has(dep) || this.vertices.has(dep))
 		) {
-			console.error("Invalid dependency detected.");
-			return "";
+			throw new Error("Invalid dependency detected.");
```
d-roak, line 131: Like the `throw` here, did you double checked everytime this function is called is not depending on the previous `return ""`?

d-roak (changes requested):
Overall the code quality is pretty poor. Let's try to make it cleaner

packages/node/src/handlers.ts
elielnfinic: If `updateMessage.vertices.length === 0`, I can't find the peerId.

packages/node/src/handlers.ts
elielnfinic: I have updated the logic, I check the name of error.

elielnfinic: I have changed to `callback` approach.

packages/object/src/hashgraph/index.ts
elielnfinic: Have removed this.

packages/node/src/handlers.ts
elielnfinic: Now using async.

packages/node/src/handlers.ts
```diff
@@ -71,14 +71,20 @@ export async function topologyMessagesHandler(
   data: { id: string, operations: {nonce: string, fn: string, args: string[] }[] }
   operations array doesn't contain the full remote operations array
 */
-function updateHandler(node: TopologyNode, data: Uint8Array) {
+async function updateHandler(
+	node: TopologyNode,
+	data: Uint8Array,
+	senderPeerId: string,
+) {
 	const updateMessage = NetworkPb.Update.decode(data);
```
d-roak, line 79: doesn't this has the sender node id already?

packages/node/src/handlers.ts
```diff
@@ -71,14 +71,20 @@ export async function topologyMessagesHandler(
...
 	const object = node.objectStore.get(updateMessage.objectId);
 	if (!object) {
 		console.error("topology::node::updateHandler", "Object not found");
 		return false;
 	}
 
+	await node.syncObject(updateMessage.objectId, senderPeerId);
```
d-roak, line 86: why are you syncing ALWAYS when you receive an update?

packages/node/src/handlers.ts
```diff
@@ -173,6 +180,7 @@ function syncAcceptHandler(
 	});
 
 	if (vertices.length !== 0) {
+		await node.syncObject(object.id, sender);
```
d-roak, line 183: why are u starting a sync process in the middle of a sync?

d-roak (changes requested):

d-roak (approved):

trungnotchung (approved):
lgtm

PR #253, opened 2024-11-14
Add access control list to CRO

trungnotchung:


packages/blueprints/src/AddWinsSet/index.ts
```diff
@@ -6,14 +6,17 @@ import {
...
 	type Vertex,
 } from "@topology-foundation/object";
+import { ACL } from "../AccessControlList/index.js";
 
 export class AddWinsSet<T> implements CRO {
-	operations: string[] = ["add", "remove"];
+	operations: string[] = ["add", "remove", "grant", "revoke"];
 	state: Map<T, boolean>;
 	semanticsType = SemanticsType.pair;
+	accessControlList: ACL;
```
JanLewDev, line 15: I think we decided that ACL will be enabled by default, and may be not used. So the attribute should also appear in the `interface CRO` definition

trungnotchung: currently, ACL is implementing the CRO interface. if we have the attribute in the interface CRO, will ACL have ACL?

PR #263, opened 2024-11-18
feat: add CROState to vertices; partial linearisation

### LINKED ISSUE #258: Partial linearisation
Currently the whole hashgraph is linearised, which calculates the same states many times

### LINKED ISSUE #259: Vertex state
Add state of the CRO for every vertex corresponding to the state at that point

JanLewDev:
Closes #259, closes #258


packages/object/src/index.ts
```diff
@@ -22,14 +22,18 @@ export interface CRO {
 export class CROState {
 	cro: CRO;
 	linearizedOperations: Operation[];
-	// past lightcone includes the vertex itself, root has in-degree 0
-	pastLightconeInDegree: Map<Hash, number>;
-	// ACL state in the near future
-
-	constructor(cro: CRO, linearizedOperations: Operation[] = [], pastLightconeInDegree: Map<Hash, number> = new Map()) {
+	// the past includes all vertices that the vertex depends on and the vertex itself
+	// root has in-degree 0
```
d-roak, line 26: out-degree?

JanLewDev: It's the in-degree when looking at forward edges, which we are traversing from the root in `topoSortPast`

packages/object/src/index.ts
```diff
@@ -18,6 +19,24 @@ export interface CRO {
 	mergeCallback: (operations: Operation[]) => void;
 }
 
+export class CROState {
+	cro: CRO;
+	linearizedOperations: Operation[];
+	// the past includes all vertices that the vertex depends on and the vertex itself
+	// root has in-degree 0
+	pastInDegree: Map<Hash, number>;
```
trungnotchung, line 27: why don't we store the topological sorted vertices instead of pastInDegree?

packages/object/src/index.ts
```diff
@@ -18,6 +19,24 @@ export interface CRO {
...
+	cro: CRO;
+	linearizedOperations: Operation[];
+	// the past includes all vertices that the vertex depends on and the vertex itself
+	// root has in-degree 0
+	past: Set<Hash>;
+
+	constructor(
+		cro: CRO,
+		linearizedOperations: Operation[] = [],
+		past: Set<Hash> = new Set(),
```
d-roak, line 32:
```suggestion
		linearizedOperations: Operation[],
		past: Set<Hash>,
```

there's no case that the constructor should be passed without this inputs

packages/object/src/index.ts
d-roak: requires interpretation. Why is it useful to have it in the comments?

packages/object/src/index.ts
```diff
@@ -87,6 +109,21 @@ export class TopologyObject implements ITopologyObject {
...
 		const vertex = this.hashGraph.addToFrontier({ type: fn, value: args });
+		// the vertex certainly has dependencies
+		let past: Set<Hash> =
+			this.states.get(vertex.dependencies[0])?.past || new Set();
+		for (const dep of vertex.dependencies) {
+			past = new Set([
+				...past,
+				...(this.states.get(dep)?.past || new Set<string>()),
+			]);
+		}
```
d-roak, line 120: the first get is redundant. You are going to iterate it anyway on the loop

packages/object/src/index.ts
```diff
@@ -87,6 +109,21 @@ export class TopologyObject implements ITopologyObject {
 	// biome-ignore lint: value can't be unknown because of protobuf
 	callFn(fn: string, args: any) {
 		const vertex = this.hashGraph.addToFrontier({ type: fn, value: args });
+		// the vertex certainly has dependencies
+		let past: Set<Hash> =
+			this.states.get(vertex.dependencies[0])?.past || new Set();
+		for (const dep of vertex.dependencies) {
+			past = new Set([
+				...past,
+				...(this.states.get(dep)?.past || new Set<string>()),
```
d-roak, line 118: `this.states.get(dep)?.past` is only undefined if `get()` fails to return. It should be checked beforehand, it shouldn't execute without prompting an error

packages/object/src/linearize/pairSemantics.ts
```diff
@@ -1,12 +1,16 @@
...
 	type HashGraph,
 	type Operation,
 } from "../hashgraph/index.js";
 
-export function linearizePair(hashGraph: HashGraph): Operation[] {
-	const order = hashGraph.topologicalSort(true);
-	const result: Operation[] = [];
+export function linearizePair(
+	hashGraph: HashGraph,
+	subgraph?: Set<Hash>,
```
d-roak, line 10: not consistent

d-roak (changes requested):

packages/object/src/hashgraph/index.ts
```diff
@@ -306,7 +312,7 @@ export class HashGraph {
 		}
 
 		const visited = new Map<Hash, number>();
-		this.depthFirstSearch(visited);
+		this.depthFirstSearch(undefined, visited);
```
trungnotchung, line 315: could you change "past" to an optional parameter so that it's not required to be passed here?

packages/object/src/index.ts
```diff
@@ -87,6 +109,21 @@ export class TopologyObject implements ITopologyObject {
 	// biome-ignore lint: value can't be unknown because of protobuf
 	callFn(fn: string, args: any) {
 		const vertex = this.hashGraph.addToFrontier({ type: fn, value: args });
+		// the vertex certainly has dependencies
+		let past: Set<Hash> =
+			this.states.get(vertex.dependencies[0])?.past || new Set();
```
trungnotchung, line 114:
```suggestion
		let past: Set<Hash> = new Set();
```
it can help make the logic less complex

packages/object/src/index.ts
```diff
@@ -87,6 +109,21 @@ export class TopologyObject implements ITopologyObject {
...
+		for (const dep of vertex.dependencies) {
+			past = new Set([
+				...past,
+				...(this.states.get(dep)?.past || new Set<string>()),
+			]);
+		}
+		past.add(vertex.hash);
+		this.states.set(
+			vertex.hash,
+			new CROState(this.cro as CRO, this.hashGraph.linearizeOperations(), past),
```
trungnotchung, line 124: this cro will be a reference

packages/object/src/index.ts
```diff
@@ -115,6 +152,24 @@ export class TopologyObject implements ITopologyObject {
...
+				}
+				past.add(vertex.hash);
+				this.states.set(
+					vertex.hash,
+					new CROState(
+						this.cro as CRO,
+						this.hashGraph.linearizeOperations(past),
+						past,
+					),
+				);
```
trungnotchung, line 172: this code is duplicate, you should write a function to do it

packages/object/src/linearize/multipleSemantics.ts
```diff
@@ -111,3 +112,99 @@ export function linearizeMultiple(hashGraph: HashGraph): Operation[] {
...
+
+export function linearizeMultipleSemantics(
+	hashGraph: HashGraph,
+	origin: Hash,
+	subgraph: Set<string>,
+): Operation[] {
+	if (subgraph.size === hashGraph.vertices.size) {
+		return fullLinearization(hashGraph);
+	}
+	return partialLinearization(hashGraph, origin, subgraph);
```
JanLewDev, line 209: What the purpose of separating? full and partial only differ at the top few lines

trungnotchung: updated

packages/object/tests/hashgraph.test.ts
```diff
@@ -1,7 +1,8 @@
 import { beforeEach, describe, expect, test } from "vitest";
 import { AddWinsSet } from "../../blueprints/src/AddWinsSet/index.js";
 import { PseudoRandomWinsSet } from "../../blueprints/src/PseudoRandomWinsSet/index.js";
-import { type Operation, OperationType, TopologyObject } from "../src/index.js";
+import { CROState, type Operation, OperationType, TopologyObject } from "../src/index.js";
+import exp from "constants";
```
JanLewDev, line 5: is this used?

trungnotchung: no, i removed them

packages/object/src/index.ts
```diff
@@ -16,6 +17,18 @@ export interface CRO {
...
+
+export class CROState {
+	// biome-ignore lint: attributes can be anything
+	state: Map<string, any>;
+
+	// biome-ignore lint: attributes can be anything
+	constructor(state: Map<string, any>) {
+		this.state = state;
+	}
 }
```
d-roak, line 32: why doing this instead of creating a type for CROState?

packages/object/src/index.ts
```diff
@@ -39,6 +52,9 @@ export class TopologyObject implements ITopologyObject {
 	cro: ProxyHandler<CRO> | null;
 	hashGraph: HashGraph;
 	subscriptions: TopologyObjectCallback[];
+	// mapping from vertex hash to the CRO state
+	states: Map<string, CROState>;
+	originalCRO: CRO;
```
d-roak, line 57: why do we need the original cro? can't we copy from the `cro` instance?

packages/object/tests/hashgraph.test.ts
d-roak, line 1: revert the comments on this file

packages/object/tests/hashgraph.test.ts
```diff
@@ -468,3 +468,148 @@ describe("HashGraph for undefined operations tests", () => {
...
+	});
+
+	test("Test: Tricky merging", () => {
+		/*
+				  A1 \
+				/	  A4 \
+		root ---> B2 /\	  A6
+				\	  C5 /
+				  C3 /
+		*/
```
d-roak, line 518: format the comments to be readable on github

Issue #282, opened 2024-12-10
Add access control list


trungnotchung:
Add access control list to the DRP
* The DRP constructor should accept a list of node IDs that form the admin set
* Admin nodes can grant or revoke write permissions to/from other nodes.
* Admins cannot grant or revoke admin permissions.
* Only nodes with write permission can generate vertices.

PR #283, opened 2024-12-10
feat: add access control list to DRP

### LINKED ISSUE #289: Add vertex signature
Add a digital signature to each vertex to verify its creator when sending and receiving.

* Sending: Sign the vertex's operation data using the creator node's private key and attach the signature.
* Receiving: Verify the signature using the node's public key to ensure authenticity and integrity.

trungnotchung:
closes: #282 , closes #289


packages/node/src/handlers.ts
```diff
@@ -218,3 +213,84 @@ export function drpObjectChangesHandler(
...
+			return;
+		}
+
+		vertex.signature = await node.networkNode.signVertexOperation(vertex);
+	});
+
+	await Promise.all(signPromises);
+}
+
+export async function verifyIncomingVertices(
```
anhnd350309, line 229: I think we need to test this function. It contains a lot of logic, including: checking if acl is undefined, checking if acl exists, and validating the vertices.

packages/blueprints/src/AccessControl/index.ts
```diff
@@ -0,0 +1,88 @@
...
+	isWriter(peerId: string): boolean {
+		return this._writers.has(peerId);
+	}
+
+	getPeerKey(peerId: string): string | undefined {
+		return this.peerKeyStore.get(peerId);
+	}
+
+	resolveConflicts(vertices: Vertex[]): ResolveConflictsType {
+		if (!vertices[0].operation || !vertices[1].operation)
```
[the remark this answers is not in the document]
trungnotchung: it always has 2 items

packages/node/src/handlers.ts
trungnotchung: i added some tests

anhnd350309 (approved):

magnified103 (approved):

JanLewDev (approved):

packages/blueprints/src/AddWinsSetWithACL/index.ts
```diff
@@ -0,0 +1,116 @@
...
+		return this.state.get(value) === true;
+	}
+
+	isAdmin(peerId: string): boolean | undefined {
+		return this.acl?.isAdmin(peerId);
+	}
+
+	isWriter(peerId: string): boolean | undefined {
+		return this.acl?.isWriter(peerId);
+	}
```
d-roak, line 81: why do we need this methods here? the `acl` is exposed, you can access this through `addWins.acl.isAdmin(peerId)`

packages/blueprints/src/AddWinsSetWithACL/index.ts
```diff
@@ -0,0 +1,116 @@
...
+	type ResolveConflictsType,
+	SemanticsType,
+	type Vertex,
+} from "@ts-drp/object";
+import { ACL } from "../AccessControl/index.js";
+
+export class AddWinsSetWithACL<T> implements DRP {
+	operations: string[] = ["add", "remove"];
+	state: Map<T, boolean>;
+	acl?: IACL & DRP;
```
d-roak, line 14: this can't be undefined, you should remove the `?`

packages/blueprints/src/AccessControl/index.ts
```diff
@@ -0,0 +1,88 @@
+import {
+	ActionType,
+	type DRP,
+	type IACL,
+	type ResolveConflictsType,
+	SemanticsType,
+	type Vertex,
+} from "@ts-drp/object";
+
+export enum AccessControlConflictResolution {
```
d-roak, line 10:
```suggestion
export enum ACLConflictResolution {
```

packages/blueprints/src/AccessControl/index.ts
```diff
@@ -0,0 +1,88 @@
...
+	GrantWins = 0,
+	RevokeWins = 1,
+}
+
+export class ACL implements IACL, DRP {
+	operations: string[] = ["grant", "revoke"];
+	semanticsType = SemanticsType.pair;
+	peerKeyStore = new Map<string, string>();
+
+	private _conflictResolution: AccessControlConflictResolution;
```
d-roak, line 20:
```suggestion
	private _conflictResolution: ACLConflictResolution;
```

packages/blueprints/src/AccessControl/index.ts
```diff
@@ -0,0 +1,88 @@
...
+	private _admins: Set<string>;
+	private _writers: Set<string>;
+
+	constructor(
+		admins: Map<string, string>,
+		conflictResolution?: AccessControlConflictResolution,
```
d-roak, line 26:
```suggestion
		conflictResolution?: ACLConflictResolution,
```

packages/blueprints/src/AccessControl/index.ts
```diff
@@ -0,0 +1,88 @@
...
+	) {
+		this._admins = new Set(admins.keys());
+		this._writers = new Set(admins.keys());
```
d-roak, line 29: why not store both as maps and drop the peerKeyStore?
to check the contains it's faster in the map and if you want the set you can just go with `Object.keys(this._admins)`

packages/blueprints/src/AccessControl/index.ts
```diff
@@ -0,0 +1,88 @@
...
+		this.peerKeyStore = admins;
+		this._conflictResolution =
+			conflictResolution ?? AccessControlConflictResolution.RevokeWins;
```
d-roak, line 32:
```suggestion
			conflictResolution ?? ACLConflictResolution.RevokeWins;
```

packages/blueprints/src/AddWinsSetWithACL/index.ts
```diff
@@ -0,0 +1,116 @@
...
+		if (!this.acl.isAdmin(sender)) {
+			throw new Error("Only admins can revoke access.");
+		}
+		if (this.acl.isAdmin(target)) {
+			throw new Error(
+				"Cannot revoke permissions from a node with admin privileges.",
+			);
+		}
+		this.acl.revoke(target);
+	}
```
d-roak, line 69: why do we need this methods here? the `acl` is exposed, you can access this through `addWins.acl.grant(...)`

packages/network/src/node.ts
```diff
@@ -61,13 +65,13 @@ export class DRPNetworkNode {
...
 	async start() {
-		let privateKey = undefined;
 		if (this._config?.private_key_seed) {
 			const tmp = this._config.private_key_seed.padEnd(32, "0");
-			privateKey = await generateKeyPairFromSeed(
+			this._privateKey = await generateKeyPairFromSeed(
 				"Ed25519",
 				uint8ArrayFromString(tmp),
 			);
+			this.publicKey = fromByteArray(this._privateKey.publicKey.raw);
```
d-roak, line 74: this was done to use a seed as a private key. you shouldn't assign here, but after the libp2p starts. There we can ensure that privateKey will have a value

d-roak, line 74: why are we using base64 and introducing a new dep? doesn't libp2p provides a to string method for keys?

packages/network/src/node.ts
```diff
@@ -288,4 +292,20 @@ export class DRPNetworkNode {
 	addCustomMessageHandler(protocol: string | string[], handler: StreamHandler) {
 		this._node?.handle(protocol, handler);
 	}
+
+	async signVertexOperation(vertex: Vertex): Promise<string> {
```
d-roak, line 296: not sure if vertex logic should be in the `network` logic. Why can't we do this logic on the `node`?

packages/node/src/handlers.ts
```diff
@@ -98,6 +93,8 @@ function syncHandler(node: DRPNode, sender: string, data: Uint8Array) {
 		return;
 	}
 
+	await signGeneratedVertices(node, object.vertices);
```
d-roak, line 96: why are we signing on sync? vertices are supposed to be signed only once at generation

packages/node/src/handlers.ts
```diff
@@ -130,31 +127,29 @@ function syncHandler(node: DRPNode, sender: string, data: Uint8Array) {
...
+	);
 
-	if (vertices.length !== 0) {
-		object.merge(vertices);
+	if (verifiedVertices.length !== 0) {
+		object.merge(verifiedVertices);
 		node.objectStore.put(object.id, object);
 	}
 
+	await signGeneratedVertices(node, object.vertices);
```
d-roak, line 152: why are we signing on sync? vertices are supposed to be signed only once at generation

packages/object/src/index.ts
```diff
@@ -14,10 +14,19 @@ import { ObjectSet } from "./utils/objectSet.js";
...
+	grant: (peerId: string, publicKey: string) => void;
+	revoke: (peerId: string) => void;
+	getPeerKey: (peerId: string) => string | undefined;
+}
+
 export interface DRP {
 	operations: string[];
 	semanticsType: SemanticsType;
 	resolveConflicts: (vertices: Vertex[]) => ResolveConflictsType;
+	acl?: IACL & DRP;
```
d-roak, line 29: why do we need to define that the acl has the `DRP` interface?

d-roak:
image

also the name of the folder should be ACL

packages/blueprints/src/AddWinsSetWithACL/index.ts
trungnotchung: the acl is optional, why this can't be undefined?

packages/blueprints/src/AddWinsSetWithACL/index.ts
trungnotchung: i will remove it

packages/network/src/node.ts
trungnotchung: because this is where i store the private key. any suggestion where i should i implement it on the `node`?

trungnotchung: inside the DRPNode?

d-roak (approved):

PR #285, opened 2024-12-10
fix: as value is an array we need to spread it otherwise the function being call will just receive an array

sfroment:


packages/object/src/index.ts
```diff
@@ -209,10 +209,14 @@ export class DRPObject implements IDRPObject {
 		}
 
 		for (const op of linearizedOperations) {
-			drp[op.type](op.value);
+			const args = Array.isArray(op.value) ? op.value : [op.value];
+			drp[op.type](...args)
```
JanLewDev, line 213:
```suggestion
			drp[op.type](...args);
```

JanLewDev (approved):
semicolon and lgtm

packages/object/src/index.ts
sfroment: if I do this I have some errors in the test

d-roak (approved):

PR #287, opened 2024-12-16
feat(rpc): add reflection

winprn:


packages/node/src/rpc/index.ts
```diff
@@ -11,6 +15,9 @@ import type {
 	UnsubscribeDRPRequest,
 	UnsubscribeDRPResponse,
 } from "../proto/drp/node/v1/rpc_pb.js";
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+const PROTO_PATH = path.resolve(__dirname, "../proto/topology/node/rpc.proto");
```
d-roak, line 20:
```suggestion
const protoPath = path.resolve(dirname(fileURLToPath(import.meta.url)), "../proto/topology/node/rpc.proto");
```

move this inside the init function

packages/node/src/rpc/index.ts
```diff
@@ -70,7 +77,11 @@ export function init(node: DRPNode) {
 		callback(null, response);
 	}
 
+	const packageDefinition = protoLoader.loadSync(PROTO_PATH);
+	const reflectionService = new reflection.ReflectionService(packageDefinition);
+
 	const server = new grpc.Server();
+	reflectionService.addToServer(server);
```
d-roak, line 84:
```suggestion
	server.addService(reflectionService);
```

does something like this work? to keep consistency

d-roak: let it be then

d-roak (approved):
lgtm

PR #288, opened 2024-12-16
feat: add timestamp to vertex

magnified103:


JanLewDev (dismissed):
If we're not differentiating between the first bootstrap sync and a later active sync, lgtm

packages/object/src/hashgraph/index.ts
```diff
@@ -145,7 +147,12 @@ export class HashGraph {
 	 * If the vertex already exists, return the hash of the existing vertex.
 	 * Throws an error if any of the dependencies are not present in the hashgraph.
 	 */
-	addVertex(operation: Operation, deps: Hash[], nodeId: string): Hash {
+	addVertex(
+		operation: Operation,
+		deps: Hash[],
+		nodeId: string,
+		timestamp?: number,
```
trungnotchung, line 154: why is this variable optional?

packages/node/src/handlers.ts
```diff
@@ -218,3 +203,29 @@ export function drpObjectChangesHandler(
 			log.error("::createObject: Invalid origin function");
 	}
 }
+
+export function cleanIncomingVertices(
```
trungnotchung, line 207: can you use `verifyIncomingVertices` to name this function? i will also use this function to verify the vertex signature

packages/object/src/hashgraph/index.ts
magnified103: If that argument is undefined then the timestamp will be set by `Date.now()`. I'm unsure whether it should be moved to the outer scope or not.

trungnotchung: because you always need to set this value, there's no reason to make this variable optional

magnified103: Changed it.

packages/node/src/handlers.ts
magnified103: I removed this function.

packages/object/src/hashgraph/index.ts
```diff
@@ -145,23 +147,40 @@ export class HashGraph {
...
-		if (
-			!deps.every((dep) => this.forwardEdges.has(dep) || this.vertices.has(dep))
-		) {
-			throw new Error("Invalid dependency detected.");
+		for (const dep of deps) {
+			const vertex = this.vertices.get(dep);
+			if (vertex === undefined) {
+				throw new Error("Invalid dependency detected.");
+			}
+			if (vertex.timestamp > timestamp) {
```
trungnotchung, line 166: i think it shoud be >=

packages/object/tests/hashgraph.test.ts
```diff
@@ -617,3 +619,66 @@ describe("Vertex state tests", () => {
...
+				Number.POSITIVE_INFINITY,
+			),
+		).toThrowError("Invalid timestamp detected.");
+	});
+
+	test("Test: Vertex's timestamp must not be less than any of its dependencies' timestamps", () => {
+		/*
+		        __ V1:ADD(1) __
+		       /               \  
+		  ROOT -- V2:ADD(2) ---- V1:ADD(4) (invalid)
```
trungnotchung, line 656:
```suggestion
		  ROOT -- V2:ADD(2) ---- V4:ADD(4) (invalid)
```

packages/object/src/hashgraph/index.ts
magnified103: Due to the clock precision, in 1ms you could add tens of vertices, and they'll have the same timestamp

JanLewDev (approved):
lgtm

packages/object/src/hashgraph/index.ts
```diff
@@ -101,13 +102,15 @@ export class HashGraph {
 
 	addToFrontier(operation: Operation): Vertex {
 		const deps = this.getFrontier();
-		const hash = computeHash(this.nodeId, operation, deps);
+		const currentTimestamp = Date.now();
+		const hash = computeHash(this.nodeId, operation, deps, currentTimestamp);
```
d-roak, line 106: Change the hardcoded root to consider a timestamp of -1

magnified103: Fixed.

packages/object/src/hashgraph/index.ts
```diff
@@ -59,12 +59,13 @@ export class HashGraph {
...
-		{ type: OperationType.NOP },
+		{ type: OperationType.NOP, value: null },
 		[],
-	)
+		-1,
+	);
 	*/
 	static readonly rootHash: Hash =
-		"02465e287e3d086f12c6edd856953ca5ad0f01d6707bf8e410b4a601314c1ca5";
+		"7ca74226e2670cd08cadd5ed351fb207ff97d505cf97051abd8c35278c93a7c8";
```
d-roak, line 68: merge `main` and recompute the hash, i think it got different because we renamed `nodeId` to `peerId`

packages/object/src/hashgraph/index.ts
```diff
@@ -151,24 +155,37 @@ export class HashGraph {
...
-		if (
-			!deps.every((dep) => this.forwardEdges.has(dep) || this.vertices.has(dep))
-		) {
-			throw new Error("Invalid dependency detected.");
+		for (const dep of deps) {
+			const vertex = this.vertices.get(dep);
+			if (vertex === undefined) {
+				throw new Error("Invalid dependency detected.");
+			}
+			if (vertex.timestamp > timestamp) {
```
d-roak, line 171:
```suggestion
			if (vertex.timestamp >= timestamp) {
```

d-roak (commented):
after both changes are applied, i'm good to merge it. If you still want to keep the discussion about clock precision. Let's continue it on discord

d-roak:
also seems there are a few conflicts, please fix those

packages/object/src/hashgraph/index.ts
magnified103: fixed.

d-roak (approved):

PR #296, opened 2024-12-23
fix: deep cloning state and original drp

### LINKED ISSUE #293: Fix drp cloning
Currently canvas example does not run, when trying to paint a pixel it throws an error (screenshot below).
image

This is because when deep cloning the drp with the current approach
```ts
		this.originalDRP = Object.create(
			Object.getPrototypeOf(drp),
			Object.getOwnPropertyDescriptors(structuredClone(drp)),
		);
```
comparing `drp` and `originalDRP`, the cloning converts Pixel objects to just objects with the functions, both logged below
image

JanLewDev:
Introduce using `cloneDeep` from `es-toolkit` to deeply clone nested objects with all functions, which makes canvas example work again and closes #293

### DIFF
--- a/packages/object/src/index.ts
+++ b/packages/object/src/index.ts
@@ -13,7 +13,7 @@ import { ObjectSet } from "./utils/objectSet.js";
 
 export * as ObjectPb from "./proto/drp/object/v1/object_pb.js";
 export * from "./hashgraph/index.js";
-import { clone, cloneDeep } from "es-toolkit";
+import { cloneDeep } from "es-toolkit";
 
 export interface IACL {
 	isWriter: (peerId: string) => boolean;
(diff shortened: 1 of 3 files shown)


packages/object/src/index.ts
```diff
@@ -13,6 +13,7 @@ import { ObjectSet } from "./utils/objectSet.js";
 
 export * as ObjectPb from "./proto/drp/object/v1/object_pb.js";
 export * from "./hashgraph/index.js";
+import { clone, cloneDeep } from "es-toolkit";
```
trungnotchung, line 16:
```suggestion
import { cloneDeep } from "es-toolkit";
```
remove unused import

d-roak (approved):
Pre-approving
Apply trung comment

trungnotchung (approved):
lgtm

anhnd350309 (approved):

PR #300, opened 2024-12-27
fix: sign vertex hash instead of vertex operation

### LINKED ISSUE #299: Sign vertex hash instead of vertex operation
Currently, the node signs only the vertex operation, which is insufficient as the vertex contains additional information, and the operation is already hashed in the vertex hash: https://github.com/topology-foundation/ts-drp/blob/4bce724dbea2486a36965a55a7818225da04f075/packages/node/src/handlers.ts#L222

Update the implementation to sign the vertex hash instead of the vertex operation to ensure all vertex data is covered.

trungnotchung:
Updated the node to sign the vertex hash instead of the vertex operation. This ensures all vertex data, including the hashed operation, is covered. Adjusted tests to reflect the change and verified functionality. Closes #299


d-roak (approved):
lgtm

PR #303, opened 2024-12-30
feat: check writer permission before merging

### LINKED ISSUE #302: Check writer permission before merging
Currently, we apply operations after checking its signature. With the ACL, we need to check if the peer that generated each vertex has permission to write or not before merging:
* If the ACL exists, check permission with `drp.acl.isWriter(peerId)`.
* If the ACL does not exist, blind merging.

trungnotchung:
Implemented a writer permission check before merging vertices.

* If ACL exists, verify permissions using drp.acl.isWriter(peerId).
* If there is no ACL, proceed with blind merging.

Update AddWinsSetWithACL because there is no need to check writer permission in the ACL. Updated tests to validate both scenarios. Resolve #302

### DIFF
--- a/packages/object/src/index.ts
+++ b/packages/object/src/index.ts
@@ -165,7 +165,7 @@ export class DRPObject implements IDRPObject {
 
 			try {
 				const drp = this._computeDRP(vertex.dependencies);
-				if (drp.acl && !drp.acl.isWriter(vertex.peerId)) {
+				if (!this._checkWriterPermission(drp, vertex.peerId)) {
 					throw new Error(`${vertex.peerId} does not have write permission.`);
 				}
 
@@ -200,6 +200,13 @@ export class DRPObject implements IDRPObject {
 		for (const callback of this.subscriptions) {
 			callback(this, origin, vertices);
 		}
+	}
+
+	private _checkWriterPermission(drp: DRP, peerId: string): boolean {
+		if (drp.acl) {
+			return drp.acl.isWriter(peerId);
+		}
+		return true;
 	}
 
 	private _applyOperation(drp: DRP, operation: Operation) {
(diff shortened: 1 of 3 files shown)


packages/object/src/index.ts
```diff
@@ -164,6 +164,11 @@ export class DRPObject implements IDRPObject {
 			}
 
 			try {
+				const drp = this._computeDRP(vertex.dependencies);
+				if (drp.acl && !drp.acl.isWriter(vertex.peerId)) {
+					throw new Error(`${vertex.peerId} does not have write permission.`);
+				}
```
0xaldric, line 170:
```suggestion
Make another function call _checkWritePermission(drp, peerId)
```

trungnotchung: updated

anhnd350309 (approved):
lgtm

JanLewDev (approved):
lgtm

packages/object/src/index.ts
```diff
@@ -265,17 +284,32 @@ export class DRPObject implements IDRPObject {
 		return newState;
 	}
 
-	private _setState(vertex: Vertex) {
-		const newState = this._computeState(vertex.dependencies, vertex.operation);
-		this.states.set(vertex.hash, { state: newState });
+	private _computeDRPState(
+		vertexDependencies: Hash[],
+		vertexOperation?: Operation | undefined,
```
d-roak, line 289:
```suggestion
		vertexOperation?: Operation,
```

packages/object/src/index.ts
```diff
@@ -256,6 +268,13 @@ export class DRPObject implements IDRPObject {
 			this._applyOperation(drp, vertexOperation);
 		}
 
+		return drp;
+	}
+
+	private _getDRPState(
+		drp: DRP,
+		// biome-ignore lint: values can be anything
+	): Map<string, any> {
```
d-roak, line 277: why are we not using the type that it's created? `DRPState`

d-roak (approved):

PR #314, opened 2025-01-03
feat: finality with bls signature approach

magnified103:
This PR introduces:
1. `DRPCredentialStore` (node/store/index.ts) holds the private keys and public keys for vertex signing and voting.
- `DRPCredentialConfig` contains the initial seed for key generation.
- `DRPPublicCredential` contains the public keys, base64-encoded.
```
export interface DRPPublicCredential {
	ed25519PublicKey: string;
	blsPublicKey: string;
}
```
2. `FinalityState` and `FinalityStore` (object/finality) hold the finality and voting status for hashgraph vertices.
- DEFAULT_FINALITY_THRESHOLD equals 0.51
3. BLS signature collection scheme:
- `Attestation` object holds a vertex hash and a signature representing a single vote.
```
message Attestation {
  string data = 1;
  bytes signature = 2;
}
```
- `AggregatedAttestation` contains a vertex hash, an aggregate of multiple signatures and a bitset indicating who voted.
```
message AggregatedAttestation {
  string data = 1;
  bytes signature = 2;
  bytes aggregation_bits = 3;
}
```
- `ATTESTATION_UPDATE` broadcast message: broadcast the attestations generated by a node upon receiving an UPDATE message.
```
message AttestationUpdate {
  string object_id = 1;
  repeated drp.object.v1.Attestation attestations = 2;
}
```


packages/network/src/node.ts
```diff
@@ -66,19 +62,14 @@ export class DRPNetworkNode {
 	}
 
 	async start() {
+		let privateKey = undefined;
 		if (this._config?.private_key_seed) {
 			const tmp = this._config.private_key_seed.padEnd(32, "0");
```
winprn, line 67:
```suggestion
			const seed = this._config.private_key_seed.padEnd(32, "0");
```

packages/object/src/finality/index.ts
```diff
@@ -0,0 +1,194 @@
...
+		this.voterCredentials = peerIds.map((peerId) =>
+			voters.get(peerId),
+		) as DRPPublicCredential[];
+
+		this.voterIndices = new Map();
+		for (let i = 0; i < peerIds.length; i++) {
+			this.voterIndices.set(peerIds[i], i);
+		}
+
+		this.aggregation_bits = new BitSet((peerIds.length + 31) >> 5);
```
winprn, line 39: what are these numbers' meaning?

packages/object/src/finality/index.ts
```diff
@@ -0,0 +1,194 @@
...
+	voterIndices: Map<string, number>;
+	aggregation_bits: BitSet;
+	signature?: Uint8Array;
+	numberOfVotes: number;
+
+	constructor(hash: Hash, voters: Map<string, DRPPublicCredential>) {
+		this.data = hash;
+
+		// deterministic order
+		const peerIds = Array.from(voters.keys()).sort();
```
winprn, line 29: why do we need to sort them

packages/object/src/finality/index.ts
```diff
@@ -0,0 +1,194 @@
...
+		);
+
+		// public keys of voters who voted
+		const credentials = this.voterCredentials.filter((_, i) =>
+			aggregation_bits.get(i),
+		);
+
+		// verify signature validity
+		if (
+			!(await bls.asyncVerifyAggregate(
```
winprn, line 102: i think we should split it in two separated statements using a new variable

winprn (changes requested):

packages/object/src/finality/index.ts
```diff
@@ -0,0 +1,194 @@
...
+					),
+					uint8ArrayFromString(this.data),
+					signature,
+				)
+			) {
+				throw new Error("Invalid signature");
+			}
+		}
+
+		if (!this.aggregation_bits.get(index)) {
```
hoangquocvietuet, line 70: Check before, no need to check again

packages/object/src/proto/drp/object/v1/object.proto
```diff
@@ -18,6 +18,17 @@ message Vertex {
   string signature = 6;
 }
 
+message Attestation {
+  string data = 1;
```
hoangquocvietuet, line 22: Should we rename `data` to `hash`?

packages/object/src/finality/index.ts
hoangquocvietuet: Yeah i want to ask about how it works too

packages/object/src/finality/index.ts
hoangquocvietuet: Yeah i want to ask about how it works

hoangquocvietuet:
Also, you forgot to update the data type in the test case for hashgraph.

packages/object/src/finality/index.ts
magnified103: fixed #349

packages/object/src/finality/index.ts
magnified103: fixed in https://github.com/topology-foundation/ts-drp/pull/314/commits/e9ebe010ed1ceca0edadccac5108a63ba6867d29

magnified103:
> Also, you forgot to update the data type in the test case for hashgraph.

fixed in https://github.com/topology-foundation/ts-drp/pull/314/commits/6934d13591a3a7db132923f4ac5aadc95a9596bf

packages/object/src/finality/index.ts
sfroment: Why don't use verifyAggregate as here we want sync execution ?

packages/network/src/node.ts
magnified103: I reverted the code to the commit https://github.com/topology-foundation/ts-drp/blob/745320e46916514e480ee8988e97bddf22c62c5b/packages/network/src/node.ts#L66

packages/object/src/finality/index.ts
magnified103: The voter order might not be deterministic.

packages/object/src/finality/index.ts
magnified103: > i think we should split it in two separated statements using a new variable

added in https://github.com/topology-foundation/ts-drp/pull/314/commits/2e1d883b56dbec5eab4cd995748f7520b5aa41ad

packages/network/src/node.ts
winprn: so we'll leave it as it is right now?

packages/object/src/hashgraph/bitset.ts
```diff
@@ -8,9 +8,22 @@
 export class BitSet {
 	private data: Uint32Array;
 
-	constructor(bits: number) {
+	constructor(bits: number, data?: Uint8Array) {
```
sfroment, line 11: Shall we set a default size if bits is 0 ?

sfroment: let me rephrase it's more do we want a minimum size ?

magnified103: Are there any problems with empty bitsets? While empty voter sets are not expected, I don’t see any problem with allowing them.

packages/network/src/node.ts
magnified103: > so we'll leave it as it is right now?

That part was introduced in #283 . In this PR I moved it to another file, so I think it's better to revert to the previous (working) implementation.

packages/object/src/proto/drp/object/v1/object.proto
magnified103: Both are fine with me.

trungnotchung (approved):
lgtm

winprn (approved):
lgtm

hoangquocvietuet (approved):
lgtm

JanLewDev (approved):
Lgtm

packages/object/src/finality/index.ts
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	async merge(attestation: AggregatedAttestation) {
+		if (this.data !== attestation.data) {
+			throw new Error("Hash mismatch");
+		}
+
+		if (this.signature) {
+			return;
+		}
```
anhnd350309, line 82: Screenshot 2025-01-14 at 21 10 26
This code will throw an error and doesn't have any try-catch statement to catch the error-

packages/object/src/finality/index.ts
```diff
@@ -0,0 +1,183 @@
...
+		const publicKeys = this.voterCredentials
+			.filter((_, i) => aggregation_bits.get(i))
+			.map((voter) => uint8ArrayFromString(voter.blsPublicKey, "base64"));
+		const data = uint8ArrayFromString(this.data);
+
+		// verify signature validity
+		if (
+			!(await bls.asyncVerifyAggregate(publicKeys, data, attestation.signature))
+		) {
+			throw new Error("Invalid signature");
```
anhnd350309, line 99: like above, it can crash the problem if you do not try-catch statement

magnified103: fixed with https://github.com/topology-foundation/ts-drp/pull/314/commits/7c9b7ecd8b4c66de7e41525d274bed7f5c829979

packages/object/src/finality/index.ts
magnified103: I added try-catch statements in addVotes and mergeVotes (https://github.com/topology-foundation/ts-drp/pull/314/commits/7c9b7ecd8b4c66de7e41525d274bed7f5c829979)

packages/object/src/finality/index.ts
```diff
@@ -0,0 +1,183 @@
...
+		}
```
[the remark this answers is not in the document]
magnified103: changed to the sync version (https://github.com/topology-foundation/ts-drp/pull/314/commits/7c9b7ecd8b4c66de7e41525d274bed7f5c829979)

anhnd350309 (approved):
lgtm

d-roak (approved):
lgtm

PR #316, opened 2025-01-06
feat: read nodes

hoangquocvietuet:
# Description:

When a subscriber doesn't know the interface but still wants to subscribe to the DRP, we need to design a read-only object that accepts messages from other nodes about the current DRP (e.g., update messages and sync messages).

This read-only object utilizes an empty DRP to bypass many functions while retaining the `merge` function and the hashgraph's `addVertex` function.


d-roak:
name of the pr needs to follow conventional commits standard

packages/object/src/index.ts
```diff
@@ -56,6 +59,59 @@ export interface DRPObjectConfig {
...
 
+export class EmptyDRP implements DRP {
+	operations: string[] = [];
+	semanticsType: SemanticsType = SemanticsType.pair;
+	resolveConflicts(): ResolveConflictsType {
+		return { action: ActionType.Nop };
+	}
+}
+
+export class EmptyDRPObject implements IDRPObject {
```
trungnotchung, line 70: i think we should rename it to `readOnlyDRPObject`

d-roak:
closing this, as the approach to do this will highly differ from the original pr. If you think this is still useful, feel free to reopen

PR #325, opened 2025-01-08
feat: update signature type

### LINKED ISSUE #324: Change the signature type from string to uint8Array
Currently, the signature type is a string. Because of that, when storing the public key and verifying the signature, converting from string to uint8Array is required. Updating the type from string to uint8Array can avoid this and increase performance.

trungnotchung:
* Changed the signature type from string to uint8Array. 
* Updated node, nodeNetwork, and handler for these changes.
* Updated test cases and passed all tests.

Resolve #324


packages/node/src/handlers.ts
```diff
@@ -215,11 +215,17 @@ export function drpObjectChangesHandler(
...
+			vertex.signature.length !== 0
+		) {
 			return;
 		}
-
-		await node.signVertex(vertex);
+		try {
+			await node.signVertex(vertex);
+		} catch (error) {
+			log.error("Error signing vertex:", vertex.hash, error);
```
d-roak, line 227:
```suggestion
			log.error("::signGeneratedVertices: Error signing vertex:", vertex.hash, error);
```

packages/node/tests/node.test.ts
```diff
@@ -21,7 +21,10 @@ describe("DPRNode with verify and sign signature", () => {
 	beforeEach(async () => {
 		drp = new AddWinsSetWithACL(
 			new Map([
-				[drpNode.networkNode.peerId, drpNode.networkNode.publicKey || ""],
+				[
+					drpNode.networkNode.peerId,
+					drpNode.networkNode.publicKey || new Uint8Array(),
```
d-roak, line 26: although it's a test, ensure that the `networkNode.publicKey` is populated

packages/object/src/index.ts
```diff
@@ -18,9 +18,9 @@ import { cloneDeep } from "es-toolkit";
 export interface IACL {
 	isWriter: (peerId: string) => boolean;
 	isAdmin: (peerId: string) => boolean;
-	grant: (senderId: string, peerId: string, publicKey: string) => void;
+	grant: (senderId: string, peerId: string, publicKey: Uint8Array) => void;
 	revoke: (senderId: string, peerId: string) => void;
-	getPeerKey: (peerId: string) => string | undefined;
+	getPeerKey: (peerId: string) => Uint8Array | undefined;
```
d-roak, line 23: which cases are undefined here? shouldn't it be an empty Uint8Array?

d-roak (approved):
pre-approving, check the comments

packages/object/src/index.ts
trungnotchung: why should it return an empty Uint8Array instead of undefined?

packages/node/src/index.ts
```diff
@@ -101,9 +101,13 @@ export class DRPNode {
...
 		if (vertex.peerId !== this.networkNode.peerId) {
-			log.error("::signVertexOperation: Invalid peer id");
-			return "";
+			throw new Error("Invalid peerId");
+		}
+
+		try {
+			vertex.signature = await this.networkNode.sign(vertex.hash);
+		} catch (error) {
+			throw new Error(`Failed to sign vertex ${vertex.hash}: ${error}`);
```
magnified103, line 110: I think we could remove this try/catch/throw block, as the errors are already handled in the parent scope.

magnified103 (changes requested):

packages/node/src/index.ts
trungnotchung: removed

magnified103 (approved):

sfroment (approved):

d-roak (approved):

PR #327, opened 2025-01-08
feat: add development env config with local node

sfroment:
The goal of this PR is to add a way to easily be able to configure the bootstrap peers on the example.
Like that people will be able to configure the node using the .env with the mode of their choice.
Right now only the bootstrap node will be configurable but in the future we might want to have other stuff configurable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ node_modules/
 **/tsconfig.tsbuildinfo
 coverage/*
 .vscode/*
+
+# macOS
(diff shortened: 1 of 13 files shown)


d-roak:
all the examples are going to be refactored into static HTML, CSS and js files when this (https://github.com/topology-foundation/ts-drp/issues/373) gets done. I'm closing this PR as `wont fix`. Feel free to argue back and reopen

PR #337, opened 2025-01-10
fix: sign vertices before sending

### LINKED ISSUE #336: ACL: signatures in the UPDATE message are missing
Vertex signatures are not properly filled before the UPDATE broadcast, causing recipient nodes to reject those vertices.

trungnotchung:
fixes #336


d-roak (approved):
lgtm

PR #340, opened 2025-01-10
fix: ip colocation weight to 0

sfroment:
Fix colocation ip, cause there is a threshold of 10 ip colocation after that the score will start to drop


d-roak (approved):
lgtm

PR #349, opened 2025-01-13
fix: wrong initial bitset size

magnified103:
The buffer sizes in the bitsets were calculated incorrectly, causing memory consumption to be 32 times worse.

This PR fixes the issue and removes unused functions.

Memory consumption changes for `obj.hashGraph.reachablePredecessors` during linearization:
- 5k vertices: 349.9 MB -> 12.9 MB
- 2k vertices: 33.7 MB -> 1.5 MB

### DIFF
--- a/packages/object/tests/bitset.test.ts
+++ b/packages/object/tests/bitset.test.ts
@@ -6,7 +6,22 @@ describe("BitSet Test", () => {
 
 	beforeEach(() => {
 		// Bitset of size 64
-		bitset = new BitSet(2);
+		bitset = new BitSet(64);
+	});
+
+	test("Test: Bitset data", () => {
+		for (let i = 0; i < 64; i++) {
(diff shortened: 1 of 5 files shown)


packages/object/tests/bitset.test.ts
```diff
@@ -5,8 +5,23 @@ describe("BitSet Test", () => {
 	let bitset: BitSet;
 
 	beforeEach(() => {
-		// Bitset of size 64
-		bitset = new BitSet(2);
+		// Bitset of size 65
+		bitset = new BitSet(65);
```
d-roak, line 9: counter-intuitive why is it 65

magnified103: I avoid using multiples of 32 to test the Math.ceil function.

d-roak (approved):
lgtm

PR #351, opened 2025-01-13
fix: pass addr sorting to connection manager to always dial in a predeterministic way

sfroment:
This PR aim to make the dial consistent even when it's not coming from the identify.
How ? By passing the addr sorting on the connection manager side, which is invoked by calculateMultiaddrs in the libp2p code, which it self is called in dial.

### DIFF
--- a/packages/network/package.json
+++ b/packages/network/package.json
@@ -43,6 +43,7 @@
 		"@libp2p/websockets": "^9.1.1",
 		"@libp2p/webtransport": "^5.0.9",
 		"@multiformats/multiaddr": "^12.3.1",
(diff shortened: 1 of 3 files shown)


d-roak (commented):
instead of this, can we look into the issue with the native auto dialer?

https://github.com/libp2p/js-libp2p-pubsub-peer-discovery/issues/149
https://github.com/ChainSafe/js-libp2p-gossipsub/issues/510

sfroment:
Yup !

sfroment:
Ported in https://github.com/topology-foundation/ts-drp/pull/363

PR #357, opened 2025-01-14
improv: add node functions

winprn:


packages/node/src/index.ts
```diff
@@ -38,6 +38,14 @@ export class DRPNode {
 		);
 	}
 
+	async restart(config?: DRPNodeConfig): Promise<void> {
+		await this.networkNode.stop();
+		this.networkNode = new DRPNetworkNode(
+			config ? config.network_config : this.config?.network_config,
+		);
+		await this.start();
```
d-roak, line 46:
```suggestion
		await this.networkNode.restart(config ? config.network_config : this.config?.network_config);
		// handler start code
```
why not this?

winprn: in `start()` we are starting the networkNode already, this will caught the node to start twice

d-roak (approved):
lgtm

Issue #304, opened 2024-12-30
Node fails to store DRP on subscription without DRP interface


hoangquocvietuet:
**Description**:

When a node subscribes to a DRP without having prior knowledge of the DRP interface, it successfully receives messages but does not store the DRP or its associated hash graph. This leads to a scenario where the node cannot properly process or verify the DRP's data.

hoangquocvietuet:
Fixed at #399

Issue #305, opened 2024-12-30
Make ACL mandatory to all DRPs


trungnotchung:
Implement mandatory ACL enforcement across all DRPs, which is currently optional, to enhance security and prevent unauthorized access

Issue #345, opened 2025-01-11
How to resolve conflicts between ACL and DRP operations


JanLewDev:
The current resolve conflicts functions only work as intended when all operations are of the same type. 
Should we separate operations based on their type and resolve separately?
Or modify the resolution mechanism to work for a mixture of operations?

PR #368, opened 2025-01-16
improv: DRPObject creation and restructuring

d-roak:
## Description
This PR introduces a couple of changes and restructuring over the DRPObject and the ACL management
- The DRPObject initialization is done with an object instead of multiple parameters (#372)
- It's possible to create a DRPObject without a DRP instance, enabling to have HGs without having the code
- The Set blueprints were replaced for a unique SetDRP that has no conflicts
- The complex cases were removed due to the new design where functions that don't change state don't go to the HG
- ACL is now used in all DRPs (#305)
- ACL was moved from `packages/blueprints` into `packages/object`
- Separate `writers` from `finalitySigners` on the ACL (allowing for permissionless while having finality)
- `packages/object/index.js` got split into separate files to increase readability
- Restructure of the interfaces
- Restructure of the store files organization
- Remove `ACL & DRP` pattern for ACLs
- Separate the dependencies installation from the action running on tests and benchmark
- Checks if node is finality signer before signing
- Create FETCH_STATE and FETCH_STATE_RESPONSE messages to get the state for a specific vertex hash
- Create connect endpoint where external node fetch the root state for the ACL
- Serializer and deserializer for nested objects in the states

TODO (further PRs): #381 , #382 , #384


packages/node/src/operations.ts
```diff
@@ -1,36 +1,50 @@
...
+		peerId: node.networkNode.peerId,
+		id,
+		drp,
+	});
+	node.objectStore.put(id, object);
+
+	await fetchState(node, id, peerId);
+	// sync process needs to finish before subscribing
+	const retry = setInterval(async () => {
+		node.objectStore.get(id);
```
trungnotchung, line 29: this code has no effect

packages/node/src/operations.ts
```diff
@@ -1,36 +1,50 @@
...
+		if (object.acl) {
```
trungnotchung, line 30: why do we have this condition

packages/object/src/acl/index.ts
```diff
@@ -0,0 +1,140 @@
...
+		peerId: string,
+		publicKey: DRPPublicCredential,
+		group: ACLGroup,
+	): void {
+		if (!this.query_isAdmin(senderId)) {
+			throw new Error("Only admin peers can grant permissions.");
+		}
+		switch (group) {
+			case ACLGroup.Admin:
+				this._admins.set(peerId, publicKey);
```
trungnotchung, line 51: also, add writer and finality signer permission here

packages/object/src/hashgraph/index.ts
trungnotchung, line 1: can we also move enum in this file to another file

packages/object/src/index.ts
```diff
@@ -2,142 +2,152 @@ import * as crypto from "node:crypto";
...
+				.update(options.peerId)
 				.update(Math.floor(Math.random() * Number.MAX_VALUE).toString())
 				.digest("hex");
-		this.abi = abi ?? "";
-		this.bytecode = new Uint8Array();
-		this.vertices = [];
-		this.drp = drp ? new Proxy(drp, this.proxyDRPHandler(DrpType.Drp)) : null;
-		this.acl = acl ? new Proxy(acl, this.proxyDRPHandler(DrpType.Acl)) : null;
+
+		const aclObj =
```
trungnotchung, line 80: i don't like this naming, aclObj vs ObjectACL

packages/object/src/interface.ts
```diff
@@ -0,0 +1,41 @@
+import type {
+	Operation,
+	ResolveConflictsType,
+	SemanticsType,
+	Vertex,
+} from "./hashgraph/index.js";
+import type { DRPObject } from "./index.js";
+import type * as ObjectPb from "./proto/drp/object/v1/object_pb.js";
+
+export type DRPStateA = {
```
trungnotchung, line 10: unused type

packages/object/src/interface.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export interface DRP {
+	semanticsType: SemanticsType;
+	resolveConflicts: (vertices: Vertex[]) => ResolveConflictsType;
+	// biome-ignore lint: attributes can be anything
+	[key: string]: any;
+}
+
+export interface LcaAndOperations {
```
trungnotchung, line 38: can we have a better name

packages/object/src/acl/index.ts
```diff
@@ -0,0 +1,140 @@
+import {
+	ActionType,
+	type ResolveConflictsType,
+	SemanticsType,
+	type Vertex,
+} from "../index.js";
+import type { DRP, DRPPublicCredential } from "../interface.js";
+import { type ACL, ACLConflictResolution, ACLGroup } from "./interface.js";
+
+export class ObjectACL implements ACL, DRP {
```
trungnotchung, line 10:
```suggestion
export class ObjectACL implements ACL {
```

packages/node/src/operations.ts
```diff
@@ -1,36 +1,50 @@
...
+			await syncObject(node, id, peerId);
+			subscribeObject(node, id);
+			object.subscribe((obj, originFn, vertices) =>
+				drpObjectChangesHandler(node, obj, originFn, vertices),
+			);
+			clearInterval(retry);
+		}
+	}, 1000);
+	return object;
```
winprn, line 39: from my understanding, this will return the `object` first before waiting for the `interval` to be cleared, is this intended?

packages/object/tests/finality.test.ts
```diff
@@ -1,16 +1,20 @@
...
 import { BitSet } from "../src/hashgraph/bitset.js";
 import { DRPObject } from "../src/index.js";
 
 // initialize log
-const _ = new DRPObject("peer1", new AddWinsSet(), new ACL(new Map()));
+const _ = new DRPObject({
+	peerId: "peer1",
+	acl: new ObjectACL({ admins: new Map() }),
+	drp: new SetDRP(),
+});
```
winprn, line 17: what are we initializing here

packages/node/src/operations.ts
d-roak: what do you mean?

packages/object/src/acl/index.ts
d-roak: no

packages/object/src/hashgraph/index.ts
d-roak: why?

packages/object/src/interface.ts
d-roak: future pr when addressing #381, this was not introduced here

packages/node/src/operations.ts
d-roak: not intended

d-roak: intended

d-roak: reason is for you to have a DRPObject instance even if it didn't fetched the initial state of the ACL yet

packages/object/tests/finality.test.ts
d-roak: not introduced in this pr. From the comment, the logger. Check with Dung

packages/object/src/hashgraph/index.ts
d-roak: wait. which enum are you talking about?

packages/node/src/operations.ts
trungnotchung: `node.objectStore.get(id);` does not do anything

packages/object/src/acl/index.ts
trungnotchung: could you pls explain why? this node can just grant itself, and it takes more action

packages/object/src/hashgraph/index.ts
trungnotchung: `OperationType`, `ActionType`, `SemanticsType`, ...

packages/object/src/index.ts
```diff
@@ -2,142 +2,152 @@ import * as crypto from "node:crypto";
...
+				"Either publicCredential or acl must be provided to create a DRPObject",
+			);
+		}
+
+		this.peerId = options.peerId;
+		log = new Logger("drp::object", options.config?.log_config);
 		this.id =
-			id ??
+			options.id ??
 			crypto
```
hoangquocvietuet, line 74: Add this to the utility functions

packages/object/src/index.ts
```diff
@@ -2,142 +2,152 @@ import * as crypto from "node:crypto";
...
-		this.drp = drp ? new Proxy(drp, this.proxyDRPHandler(DrpType.Drp)) : null;
-		this.acl = acl ? new Proxy(acl, this.proxyDRPHandler(DrpType.Acl)) : null;
+
+		const objAcl =
+			options.acl ??
+			new ObjectACL({
+				admins: new Map([
+					[options.peerId, options.publicCredential as DRPPublicCredential],
+				]),
+				permissionless: true,
```
hoangquocvietuet, line 86: Make `permissionless` an option when creating a DRPObject

packages/object/src/index.ts
```diff
@@ -2,142 +2,152 @@ import * as crypto from "node:crypto";
...
+			this.peerId,
+			acl.resolveConflicts.bind(this.acl),
+		);
+		this.vertices = this.hashGraph.getAllVertices();
+	}
+
+	static createObject(options: {
+		peerId: string;
+		id?: string;
+		drp?: DRP;
```
hoangquocvietuet, line 130:
```suggestion
	static createObject(options: {
		peerId: string;
		id?: string;
		drp?: DRP;
		permissionless?: boolean;
```

packages/object/src/index.ts
```diff
@@ -2,142 +2,152 @@ import * as crypto from "node:crypto";
...
+			drp: options.drp,
+		});
+		return object;
+	}
+
 	resolveConflicts(vertices: Vertex[]): ResolveConflictsType {
 		if (
 			this.acl &&
-			vertices.some((v) => v.operation?.drpType === DrpType.Acl)
+			vertices.some((v) => v.operation?.drpType === DrpType.ACL)
```
hoangquocvietuet, line 148:
```suggestion
			vertices[0].operation?.drpType === DrpType.ACL
```

packages/object/src/hashgraph/index.ts
d-roak: https://github.com/topology-foundation/ts-drp/issues/376

packages/object/src/acl/index.ts
d-roak: Even if it can grant itself the roles, if he doesn't explicitly take the action to do it, it's not supposed to happen. E.g. an admin might not be supposed to sign stuff for finality

packages/node/src/index.ts
```diff
@@ -87,28 +87,55 @@ export class DRPNode {
...
+		const object = operations.connectObject(
+			this,
+			options.id,
+			options.drp,
+			options.sync?.peerId,
+		);
+		return object;
+	}
+
 	async subscribeObject(id: string) {
```
hoangquocvietuet, line 139: Delete this and replace with `connectObject` in the gRPC function at https://github.com/topology-foundation/ts-drp/blob/0c307b8981b20c63fd7113cea9bbae89d2e6d584/packages/node/src/rpc/index.ts#L28C9-L28C24.

trungnotchung (approved):
lgtm

packages/object/src/index.ts
hoangquocvietuet: We don't always have ways to fix those cases where both DRP and ACL are involved, so let's check a random index of theirs

packages/object/src/index.ts
d-roak: not introduced in this pr

packages/node/src/index.ts
hoangquocvietuet: It's for fixing https://github.com/topology-foundation/ts-drp/issues/304 ser, we need to put object into `objectStore`

packages/object/src/index.ts
d-roak: no. If you want, you can configure it with `options.acl`

winprn (approved):
lgtm

packages/object/src/index.ts
d-roak: no. The `createObject` is used for when u are connecting and you need to have an intermediary object before fetching the initial state:
- in case the original acl is permissionless: you need to have it permissionless in ur local version to run operations before fetching/syncing
- in case the original acl is permissioned and you don't have permissions: your local operations will be dropped, working correctly

packages/node/src/index.ts
d-roak: this pr won't fix #304 . It's just an intermediary step. Corrected the description

packages/object/src/index.ts
d-roak: this was not introduced here. And this needs to be refactored with #345. I won't change it in this pr

packages/object/src/utils/serializer.ts
```diff
@@ -0,0 +1,145 @@
+import { Value } from "../proto/google/protobuf/struct_pb.js";
+import { toString as uint8ArrayToString } from "uint8arrays/to-string";
```
anhnd350309, line 2: unuse import

JanLewDev (approved):
lgtm

packages/node/src/handlers.ts
```diff
@@ -46,45 +59,118 @@ export async function drpMessagesHandler(
...
+		aclState: serializeStateMessage(aclState),
+		drpState: serializeStateMessage(drpState),
+	});
+
+	const message = NetworkPb.Message.create({
+		sender: node.networkNode.peerId,
+		type: NetworkPb.MessageType.MESSAGE_TYPE_FETCH_STATE_RESPONSE,
+		data: NetworkPb.FetchStateResponse.encode(response).finish(),
+	});
+	node.networkNode.sendMessage(sender, message);
```
sfroment, line 92: Are you willingly not awaiting the send ?

packages/node/src/index.ts
```diff
@@ -87,28 +87,55 @@ export class DRPNode {
...
+			peerId: this.networkNode.peerId,
+			publicCredential: options.acl
+				? undefined
+				: this.credentialStore.getPublicCredential(),
+			acl: options.acl,
+			drp: options.drp,
+			id: options.id,
+		});
 		operations.createObject(this, object);
 		operations.subscribeObject(this, object.id);
```
sfroment, line 109: same for those 2 is that intended not to wait ?

packages/node/src/operations.ts
```diff
@@ -1,34 +1,47 @@
...
+}
+
 /* data: { id: string } */
 export async function subscribeObject(node: DRPNode, objectId: string) {
 	node.networkNode.subscribe(objectId);
-	node.networkNode.addGroupMessageHandler(objectId, async (e) =>
+	node.networkNode.addGroupMessageHandler(objectId, (e) =>
 		drpMessagesHandler(node, undefined, e.detail.msg.data),
```
sfroment, line 45:
```suggestion
		drpMessagesHandler(node, undefined, e.detail.msg.data).catch((err) => console.error("failed to subscribe to object err:", err)),
```

sfroment (commented):
It's a shame that biome doesn't have the rules for floating promise

packages/node/src/handlers.ts
d-roak: why do you need to wait here?

packages/node/src/index.ts
d-roak: same here

packages/node/src/operations.ts
d-roak: why are you trying to catch the error there?

anhnd350309 (approved):
lgtm

packages/object/src/hashgraph/index.ts
```diff
@@ -309,7 +311,7 @@ export class HashGraph {
 		origin: Hash = HashGraph.rootHash,
 		subgraph: ObjectSet<string> = new ObjectSet(this.vertices.keys()),
 	): Operation[] {
-		switch (this.semanticsType) {
+		switch (this.semanticsTypeDRP) {
```
hoangquocvietuet, line 314: When this.semanticsTypeDRP is undefined, it should return the entire hashgraph in topological order, not an empty one, in order to maintain the hashgraph without knowing the DRP

hoangquocvietuet (commented):
Add `drpType: DrpType.ACL` in `vertex` in ACL testcases

packages/object/tests/hashgraph.bench.ts
```diff
@@ -56,13 +61,12 @@ benchmarkForAddWinSet(
...
-		// biome-ignore lint/suspicious/noExplicitAny: <explanation>
-		null as any,
-	);
-	const drp = object.drp as ConflictResolvingMap<number, number>;
+	const object: DRPObject = new DRPObject({
+		peerId: "peer1",
+		acl,
+		drp: new MapDRP<number, number>(),
+	});
+	const drp = object.drp as MapDRP<number, number>;
```
hoangquocvietuet, line 69:
```suggestion
    const drpHandler = drpObject.drp as unknown as MapDRP<number, number>;
```

packages/object/tests/hashgraph.bench.ts
```diff
@@ -71,13 +75,12 @@ suite.add("Create a HashGraph with 1000 operations for set wins map", () => {
...
-			// biome-ignore lint/suspicious/noExplicitAny: <explanation>
-			null as any,
-		);
-		const drp = object.drp as ConflictResolvingMap<number, number>;
+		const object: DRPObject = new DRPObject({
+			peerId: "peer1",
+			acl,
+			drp: new MapDRP<number, number>(),
+		});
+		const drp = object.drp as MapDRP<number, number>;
```
hoangquocvietuet, line 83:
```suggestion
		const drp = object.drp as unknown as MapDRP<number, number>;
```

hoangquocvietuet (changes requested):
lgtm after merge

packages/object/src/hashgraph/index.ts
d-roak: not introduced in this pr

packages/object/tests/hashgraph.bench.ts
d-roak: non breaking, and not introduced in this pr. If you want to fix ur linter, discover the root cause (why it acts different then the set drp)

packages/object/tests/hashgraph.bench.ts
d-roak: non breaking, and not introduced in this pr. If you want to fix ur linter, discover the root cause (why it acts different then the set drp)

packages/node/src/handlers.ts
sfroment: just cause sendMessage is async, and if we do not want to await I would like a .catch in case the function throw.
Also same in sendMessage we aren't await the write on the stream, so even awaiting sendMessage isn't allowing to "really" wait for a sendMessage to happen

packages/node/src/operations.ts
sfroment: Just in case we have something that throw an error in the drpMessageHandler, right now the error will just be dropped

packages/node/src/index.ts
sfroment: so for subscribe actually we shall remove the async from the function. But for the latter 
```
		if (options.sync?.enabled) {
			operations.syncObject(this, object.id, options.sync.peerId);
		}
		```
		I would expect to wait for the sync to be completed

File: packages/node/src/operations.ts [hunk 22]
```diff
@@ -1,34 +1,47 @@
...
+				drpObjectChangesHandler(node, obj, originFn, vertices),
+			);
+			clearInterval(retry);
+		}
+	}, 1000);
+	return object;
+}
+
 /* data: { id: string } */
 export async function subscribeObject(node: DRPNode, objectId: string) {
```
sfroment, line 42:
```suggestion
export function subscribeObject(node: DRPNode, objectId: string) {
```

hoangquocvietuet (approved):

Issue #358, opened 2025-01-14
Enhance Error Handling and Argument Validation in _applyOperation Function


hoangquocvietuet:
For instance, missing the error handler: https://github.com/topology-foundation/ts-drp/blob/cc45067aec53f4148945c04a532ef6647ea21bdd/packages/object/src/index.ts#L146-L151

For validation: Verify if `(...value)` matches the arguments of `target[methodName]`.

PR #383, opened 2025-01-21
fix: add apply operation error handler

hoangquocvietuet:
Issue #358

### DIFF
--- a/packages/object/src/index.ts
+++ b/packages/object/src/index.ts
@@ -194,7 +194,9 @@ export class DRPObject implements IDRPObject {
 		try {
 			this._applyOperation(drp, { opType: fn, value: args, drpType });
 		} catch (e) {
-			log.error(`Object ${this.id} failed to apply operation: ${e} in callFn ${fn}`);
+			log.error(
+				`Object ${this.id} failed to apply operation: ${e} in callFn ${fn}`,
+			);
 			return;
 		}
 
@@ -260,7 +262,9 @@ export class DRPObject implements IDRPObject {
 					try {
 						this._applyOperation(drp, vertex.operation);
 					} catch (e) {
-						log.error(`Object ${this.id} failed to apply operation: ${e} in merge`);
+						log.error(
+							`Object ${this.id} failed to apply operation: ${e} in merge`,
+						);
 						missing.push(vertex.hash);
 						continue;
 					}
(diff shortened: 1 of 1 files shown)


anhnd350309 (approved):
lgtm

packages/object/src/index.ts
```diff
@@ -177,7 +177,7 @@ export class DRPObject implements IDRPObject {
 		};
 	}
 
-	callFn(
+	private callFn(
```
winprn, line 180: why we need it to be private?

packages/object/src/index.ts
```diff
@@ -252,7 +259,15 @@ export class DRPObject implements IDRPObject {
 						vertex.timestamp,
 						vertex.signature,
 					);
-					this._applyOperation(drp, vertex.operation);
+					try {
```
trungnotchung, line 262: why do you have a try-catch inside another

trungnotchung (commented):
add some tests for merging case

packages/object/src/index.ts
hoangquocvietuet: Oops, it handles in the same way, so the outer try-catch is enough. I've removed the inside one

packages/object/src/index.ts
hoangquocvietuet: We don’t want developers to access this function because it changes the DRP state without altering the DRP itself, as it doesn’t go through a proxy call

hoangquocvietuet:
> add some tests for merging case

Which case ser

trungnotchung:
> > add some tests for merging case
> 
> Which case ser

for the try-catch when applying operation

winprn (approved):
lgtm

trungnotchung (approved):
lgtm

packages/object/src/index.ts
```diff
@@ -191,7 +191,14 @@ export class DRPObject implements IDRPObject {
 			preOperationDRP = this._computeDRP(this.hashGraph.getFrontier());
 		}
 		const drp = cloneDeep(preOperationDRP);
-		this._applyOperation(drp, { opType: fn, value: args, drpType });
+		try {
+			this._applyOperation(drp, { opType: fn, value: args, drpType });
+		} catch (e) {
+			log.error(
+				`Object ${this.id} failed to apply operation: ${e} in callFn ${fn}`,
```
d-roak, line 198:
```suggestion
				`::drpObject::callFn: Object ${this.id} failed to apply operation: ${e} in callFn ${fn}`,
```

hoangquocvietuet: fixed

d-roak (approved):
lgtm

PR #394, opened 2025-01-22
feat: use DHT to make peer discovery

sfroment:


d-roak:
how did you tested this?

d-roak:
add description on why are you adding this. And all the assumptions on how everything works. We tried to use the dht in the past but it was not that simple

PR #415, opened 2025-01-30
fix: swap action type behaviour

### LINKED ISSUE #407: Fix swap action type behaviour
Swap action type should also swap the reachable predecessors bitsets.
Reference:
https://github.com/topology-foundation/ts-drp/blob/b7bd0bc90a5677afdfeb48f0171f5401e5218328/packages/object/src/linearize/pairSemantics.ts#L48-L51
Also, reassign the anchor when swapping.

JanLewDev:
Closes #407


d-roak (approved):
lgtm

PR #418, opened 2025-01-31
feat: wait for node to be dialable after libp2p.start

sfroment:


packages/network/src/node.ts
```diff
@@ -195,8 +195,11 @@ export class DRPNetworkNode {
 				webTransport(),
 			],
 		});
+		log.info(
+			"running on:",
```
d-roak, line 199:
```suggestion
			"::start: running on:",
```

PR #419, opened 2025-02-03
feat: add functions in network

hoangquocvietuet:
Added:
- gossip_sub_config: specifies network performance and tuning configuration parameters
- `waitForPeer(peerId: string, timeout = 5000)`: Waits for a peer with the given peerId to connect within a timeout (default: 5 seconds).
- `waitForUpgradedConnection(peerId: string, timeout = 5000)`: Waits for an upgraded connection to the specified peerId within the given timeout.
- `libp2pPeerId()`: Returns the peer ID of the local libp2p node.
- `isDialable(timeout = 5000)`: Determines if the local node can accept inbound connections
- `isSubscribed(topic: string, peerId: PeerId, timeout = 5000)`: Checks if a specific peerId is subscribed to a given topic in pubsub.

### DIFF
--- a/packages/network/src/node.ts
+++ b/packages/network/src/node.ts
@@ -1,4 +1,9 @@
-import { type GossipSub, type GossipsubMessage, gossipsub } from "@chainsafe/libp2p-gossipsub";
+import {
+	type GossipSub,
+	type GossipsubMessage,
+	GossipsubOpts,
+	gossipsub,
+} from "@chainsafe/libp2p-gossipsub";
 import {
 	type TopicScoreParams,
 	createPeerScoreParams,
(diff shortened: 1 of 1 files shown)


packages/network/src/node.ts
```diff
@@ -238,6 +251,120 @@ export class DRPNetworkNode {
...
+					this._node.removeEventListener("transport:listening", transportListeningListener);
+					clearTimeout(timeoutId);
+				}
+			};
+
+			this._node.addEventListener("transport:listening", transportListeningListener);
+		});
+	}
+
+	isSubscribed(topic: string, peerId: PeerId, timeout = 5000) {
```
sfroment, line 345: It's async as you return a promise

packages/network/src/node.ts
```diff
@@ -238,6 +251,120 @@ export class DRPNetworkNode {
...
+		});
+	}
+
+	libp2pPeerId() {
+		if (!this._node) throw new Error("Node not initialized, please run .start()");
+		return this._node.peerId;
+	}
+
+	isDialable(timeout = 5000) {
+		return new Promise((resolve) => {
```
sfroment, line 311: It's async as you return a promise

PR #421, opened 2025-02-03
feat: modify acl state

anhnd350309:
## Description
This PR consolidates `_admin`, `_finalitySigners`, `_writers`, and related variables into `_authorizedPeers` to make state of being more efficient

## todo
- [x] change acl state
- [x] add some testcase

### DIFF
--- a/packages/object/src/acl/index.ts
+++ b/packages/object/src/acl/index.ts
@@ -29,28 +29,31 @@ export class ObjectACL implements ACL {
 		this._conflictResolution = options.conflictResolution ?? ACLConflictResolution.RevokeWins;
 	}
 
-	grant(senderId: string, peerId: string, publicKey: DRPPublicCredential, group: ACLGroup): void {
+	grant(senderId: string, peerId: string, group: ACLGroup, publicKey?: DRPPublicCredential): void {
 		if (!this.query_isAdmin(senderId)) {
 			throw new Error("Only admin peers can grant permissions.");
 		}
-		const existingUser = this._authorizedPeers.get(peerId);
-		const _grant = (peerId: string, group: ACLGroup) => {
-			const user = existingUser ?? { publicKey, permissions: new Set() };
-			user.permissions.add(group);
-			this._authorizedPeers.set(peerId, user);
-		};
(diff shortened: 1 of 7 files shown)


packages/object/src/acl/index.ts
```diff
@@ -8,41 +9,52 @@ export class ObjectACL implements ACL {
...
+		if (!options.permissionless) {
+			permissions.add(ACLGroup.Writer);
+		}
+
+		for (const [key, value] of options.admins) {
+			this._authorizedPeers.set(key, {
+				publicKey: value,
+				permissions: permissions,
+			});
+		}
```
d-roak, line 32: what about?

```suggestion
		const permissions = new Set<ACLGroup>([ACLGroup.Admin, ACLGroup.Finality]);
		if (!options.permissionless) {
			permissions.add(ACLGroup.Writer);
		}
		
		this._authorizedPeers = new Map([...options.admins.entries()].map(([key, value]) => [key, { publicKey: value, permissions }]));
```

packages/object/src/acl/index.ts
```diff
@@ -62,36 +74,38 @@ export class ObjectACL implements ACL {
...
 			default:
 				throw new Error("Invalid group.");
 		}
 	}
 
 	query_getFinalitySigners(): Map<string, DRPPublicCredential> {
-		return new Map(this._finalitySigners);
+		return new Map(
+			[...this._authorizedPeers.entries()]
+				.filter(([, user]) => user.permissions.has(ACLGroup.Finality))
```
d-roak, line 90:
```suggestion
				.filter(([_, user]) => user.permissions.has(ACLGroup.Finality))
```

packages/object/src/acl/index.ts
anhnd350309: yes, same convention code. changed

packages/object/src/acl/index.ts
anhnd350309: changed

vitest.workspace.ts
d-roak, line 1: why do we need to add this file?

d-roak (approved):

PR #422, opened 2025-02-03
fix: hash validation 

### LINKED ISSUE #379: Verify vertex hash before adding to the hashgraph
Currently, we're hashing the vertex data to avoid byzantine nodes changing the data of vertices. But instead of verifying the hash value's correction, we're recomputing it again in the function `addVertex` function.

The solution is to verify the hash value before adding vertices to the hashgraph instead of recomputing it.

trungnotchung:
* validate the hash value in the drpObject when merging instead of recompute this value in the addVertex
* update the addVertex and addToFrontier logic
* update the tests with these changes
* add tests for hash validation
resolve #379


d-roak (approved):
lgtm

Issue #429, opened 2025-02-05
`_setState` Updates Both DRP State and ACL State Incorrectly


trungnotchung:
Currently, function `_setState` updates both DRP State and ACL State with parameter `drpState`

```typescript
private _setState(vertex: Vertex, drpState?: ObjectPb.DRPState) {
		const preCompute = this.computeLCA(vertex.dependencies);
		this._setObjectACLState(vertex, preCompute, drpState);
		this._setDRPState(vertex, preCompute, drpState);
}
```

This parameter is just DRP State or ACL State but not both and can lead to incorrect vertex state

hoangquocvietuet:
Close with #430 merged

PR #430, opened 2025-02-05
fix: Updates Both DRP State and ACL State Incorrectly

anhnd350309:
to close bug #429 
- fix error when `callFn` update again state of DRP and ACL

### DIFF
--- a/packages/object/src/index.ts
+++ b/packages/object/src/index.ts
@@ -215,7 +215,14 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
 		});
 		this.hashGraph.addToFrontier(vertex);
 
-		this._setState(vertex, this._getDRPState(drp));
+		const preCompute = this.computeLCA(vertex.dependencies);
+		if (drpType === DrpType.DRP) {
+			this._setObjectACLState(vertex, preCompute);
+			this._setDRPState(vertex, preCompute, this._getDRPState(drp));
+		} else {
+			this._setObjectACLState(vertex, preCompute, this._getDRPState(drp));
+			this._setDRPState(vertex, preCompute);
+		}
 		this._initializeFinalityState(vertex.hash);
 
 		this.vertices.push(vertex);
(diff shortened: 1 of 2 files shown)


d-roak (approved):
lgtm

packages/object/src/index.ts
```diff
@@ -215,7 +215,14 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
 		});
 		this.hashGraph.addToFrontier(vertex);
 
-		this._setState(vertex, this._getDRPState(drp));
+		const preCompute = this.computeLCA(vertex.dependencies);
```
trungnotchung, line 218: you don't need to compute `preCompute` here because it will be computed later

packages/object/tests/drpobject.test.ts
```diff
@@ -25,3 +34,41 @@ describe("AccessControl tests with RevokeWins resolution", () => {
...
+			const stateKeys = state.state.map((x) => x.key);
+			expect(stateKeys).toEqual(expectedKeys);
+		}
+
+		const drpStates = drpObject.drpStates.values();
+		const expectedDrpKeys = Object.keys(drpSet);
+
+		// Check that each state contains the expected keys
+		for (const state of drpStates) {
+			if (state.state.length === 0) {
```
trungnotchung, line 67: when will it happen?

packages/object/tests/drpobject.test.ts
```diff
@@ -25,3 +34,41 @@ describe("AccessControl tests with RevokeWins resolution", () => {
...
 });
+
+describe("Drp Object should be able to change state value", () => {
+	let drpObject: DRPObject;
+
+	beforeEach(async () => {
+		drpObject = new DRPObject({ peerId: "peer1", acl, drp: new SetDRP<number>() });
+	});
+
+	it("should update ACL state keys when DRP state changes", () => {
```
trungnotchung, line 45: i think we should ensure that aclStates and drpStates have a length of at least one or your condition will always satisfied

packages/object/tests/drpobject.test.ts
anhnd350309: It's the root of hashgraph, with value 0

packages/object/src/index.ts
anhnd350309: If not pre compute, It'll compute twice times

packages/object/tests/drpobject.test.ts
anhnd350309: Can you give me a suggestion, In this test case, I try to expect the key of state to be the same

packages/object/src/index.ts
trungnotchung: no it only compute once because you already had aclState or drpState

packages/object/tests/drpobject.test.ts
trungnotchung: it can not happen. we fixed it in our lastest main

packages/object/src/index.ts
anhnd350309: updated

packages/object/tests/drpobject.test.ts
anhnd350309: fixed

trungnotchung (approved):
lgtm

Issue #404, opened 2025-01-27
Fix duplicate DPR call


sfroment:
Right now since we don't return on callFn here https://github.com/topology-foundation/ts-drp/blob/226868a7716f826ccfa6e070a604169eaa11d4fe/packages/object/src/index.ts#L178-L180. 
Any DRP method will be called twice increase by x2 the computation of any DRP. 
Cause later on in the callFn we will have this https://github.com/topology-foundation/ts-drp/blob/226868a7716f826ccfa6e070a604169eaa11d4fe/packages/object/src/index.ts#L371-L375 which will call your DRP method.
Then we will call the Reflect.apply which itself will call the same function.
Instead we shall be propagating the state we already pre computed if it has changed

PR #432, opened 2025-02-05
fix: duplicate DRP call

hoangquocvietuet:
Resolved: #404


packages/object/tests/drpobject.test.ts
```diff
@@ -25,3 +27,43 @@ describe("AccessControl tests with RevokeWins resolution", () => {
 		expect(obj.drp).toBeUndefined();
 	});
 });
+
+describe("Hi", () => {
```
sfroment, line 31:
```suggestion
describe("Duplicate call detection", () => {
```

packages/object/src/index.ts
```diff
@@ -172,49 +188,47 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
...
-			if (!deepEqual(preOperationDRP[key], drp[key])) {
-				stateChanged = true;
-				break;
-			}
+			return appliedOperationResult;
 		}
 
+		const stateChanged = Object.keys(preOperationDRP).some(
+			(key) => !deepEqual(preOperationDRP[key], drp[key])
+		);
```
trungnotchung, line 216: unrelated change

packages/object/src/index.ts
```diff
@@ -172,49 +188,47 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
...
-		if (drpType === DrpType.ACL) {
-			preOperationDRP = this._computeObjectACL(this.hashGraph.getFrontier());
-		} else {
-			preOperationDRP = this._computeDRP(this.hashGraph.getFrontier());
-		}
+
+		const isACL = drpType === DrpType.ACL;
+		const computeFn = isACL ? this._computeObjectACL : this._computeDRP;
+		const vertexDependencies = this.hashGraph.getFrontier();
+		const preOperationDRP = computeFn(vertexDependencies);
```
trungnotchung, line 199: why don't you call these functions directly instead of defining a new variable `computeFn`. it just makes the code harder to understand

packages/object/src/index.ts
```diff
@@ -378,18 +394,18 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
...
 	}
 
 	// compute the DRP based on all dependencies of the current vertex using partial linearization
-	private _computeDRP(
+	private _computeDRP = (
 		vertexDependencies: Hash[],
 		preCompute?: LcaAndOperations,
 		vertexOperation?: Operation
-	): DRP {
+	): DRP => {
```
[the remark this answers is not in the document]
sfroment: Cause otherwise this isn't bind

packages/object/src/index.ts
sfroment: I don't agree with this statement, but we can still discuss about it

sfroment: it remove and if else and also remove the fact that we have to add a let var

sfroment: this way is more js idiomatic

trungnotchung: i mean: 
```typescript
const preOperationDRP = isACL ? this._computeObjectACL(vertexDependencies) :  this._computeDRP(vertexDependencies);
```

packages/object/src/index.ts
```diff
@@ -156,7 +163,16 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
...
 								return Reflect.apply(applyTarget, thisArg, args);
 							}
-							if (!callerName?.startsWith("Proxy.")) obj.callFn(fullPropKey, args, vertexType);
+							if (!callerName?.startsWith("Proxy.")) {
+								const {
+									drp: newDRP,
+									appliedOperationResult,
+									isACL,
+								} = obj.callFn(fullPropKey, args, vertexType);
+								if (!isACL) Object.assign(obj.drp as DRP, newDRP);
```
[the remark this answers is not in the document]
hoangquocvietuet: I am good with both. But with #447, we can chose a better way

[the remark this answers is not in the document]
trungnotchung: let's update it inside the callFn so you won't need to return these values

packages/object/src/index.ts
sfroment: Oh yes 👍!

packages/object/src/index.ts
```diff
@@ -156,7 +156,10 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
 							if (callerName?.startsWith("DRPObject.resolveConflicts")) {
 								return Reflect.apply(applyTarget, thisArg, args);
 							}
-							if (!callerName?.startsWith("Proxy.")) obj.callFn(fullPropKey, args, vertexType);
+							if (!callerName?.startsWith("Proxy.")) {
+								const appliedOperationResult = obj.callFn(fullPropKey, args, vertexType);
+								return appliedOperationResult;
```
trungnotchung, line 161:
```suggestion
								return obj.callFn(fullPropKey, args, vertexType);
```

sfroment (approved):

packages/object/src/index.ts
```diff
@@ -149,15 +149,17 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
 					const fullPropKey = String(propKey);
 					return new Proxy(target[propKey as keyof object], {
 						apply(applyTarget, thisArg, args) {
-							if ((propKey as string).startsWith("query_")) {
-								return Reflect.apply(applyTarget, thisArg, args);
-							}
 							const callerName = new Error().stack?.split("\n")[2]?.trim().split(" ")[1];
-							if (callerName?.startsWith("DRPObject.resolveConflicts")) {
+							const startWithQuery = (propKey as string).startsWith("query_");
```
magnified103, line 153: How does `propKey` differ from `callerName` in this context?

packages/object/src/index.ts
```diff
@@ -226,6 +227,11 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
 
 		this.vertices.push(vertex);
 		this._notify("callFn", [vertex]);
+
+		if (!isACL) Object.assign(this.drp as DRP, drp);
+		else Object.assign(this.acl as ObjectACL, drp);
```
magnified103, line 232: Why didn’t the old code include these lines?

sfroment: Cause before callFn was follow by an apply which would ultimately change the state of the obj

packages/object/src/index.ts
magnified103: Not related to this PR

PR #437, opened 2025-02-05
fix: root vertex state can be changed

### LINKED ISSUE #434: `_setRootStates` Stores ACL and DRP State as references
`_setRootStates` assigns `aclState` and `drpState` as references instead of deep copies:
```typescript
this.aclStates.set(HashGraph.rootHash, { state: aclState });
this.drpStates.set(HashGraph.rootHash, { state: drpState });
```
Since `aclState` and `drpState` reference `acl` and `drp`, updates to these objects also modify the root vertex state, causing unintended changes.

trungnotchung:
* replaced direct references with deep copies using cloneDeep
* uncommented root vertex drp state
* added tests to ensure root vertex state can not be changed
resolve #434

### DIFF
--- a/packages/object/src/index.ts
+++ b/packages/object/src/index.ts
@@ -520,18 +520,18 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
 		const aclState = [];
 		for (const key of Object.keys(acl)) {
 			if (typeof acl[key] !== "function") {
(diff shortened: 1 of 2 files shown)


hoangquocvietuet (approved):
lgtm

d-roak (approved):
lgtm

PR #440, opened 2025-02-06
improv: use only new vertices for hashgraph merging

### LINKED ISSUE #420: Hashgraph `merge` triggers callbacks with all vertices as the argument
Idealy, `_notify` should be called with the vertices that `merge` just added.
https://github.com/topology-foundation/ts-drp/blob/9b423533b2dddfd9200d30f8bf41816084ee4325/packages/object/src/index.ts#L282-L285

winprn:
resolves #420

### DIFF
--- a/packages/object/src/index.ts
+++ b/packages/object/src/index.ts
@@ -240,6 +240,7 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
 	 */
 	private _mergeWithDrp(vertices: Vertex[]): [merged: boolean, missing: string[]] {
 		const missing: Hash[] = [];
+		const newVertices: Vertex[] = this.vertices;
 		for (const vertex of vertices) {
 			// Check to avoid manually crafted `undefined` operations
 			if (!vertex.operation || this.hashGraph.vertices.has(vertex.hash)) {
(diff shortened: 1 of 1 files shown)


packages/object/src/index.ts
```diff
@@ -239,7 +239,7 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
 	/* Merges the vertices into the hashgraph using DRP
 	 */
 	private _mergeWithDrp(vertices: Vertex[]): [merged: boolean, missing: string[]] {
-		const missing = [];
+		const missing: string[] = [];
```
hoangquocvietuet, line 242:
```suggestion
		const missing: Hash[] = [];
```

winprn: forgot that we have this alias, fixed

packages/object/src/index.ts
```diff
@@ -281,9 +281,10 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
 		}
 
 		this.vertices = this.hashGraph.getAllVertices();
+		const newVertices = this.vertices.filter((v) => !missing.includes(v.hash));
```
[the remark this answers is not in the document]
winprn: updated

PR #445, opened 2025-02-07
fix: vertex not found

trungnotchung:


d-roak (approved):
lgtm

packages/object/src/index.ts
```diff
@@ -572,9 +594,9 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
 	}
 }
 
-export function computeHash(
+function computeHash(
 	peerId: string,
-	operation: Operation,
+	operation: Operation | undefined,
```
d-roak, line 599: can the operation be undefined?

anhnd350309 (approved):
lgtm

packages/object/src/index.ts
trungnotchung: it can't but the vertex.operation type forced me to do that

d-roak: add the `vertex.operation` check before calling the `computeHash(...)`, no?

trungnotchung: https://github.com/drp-tech/ts-drp/blob/93248bfdab4c4b9606c28f79cd9793af0b518bf0/packages/object/src/index.ts#L286
we already checked it. if we want it to work, i will need to move this check to function `validateVertex`

packages/object/src/hashgraph/index.ts
```diff
@@ -137,30 +137,6 @@ export class HashGraph {
 	 * Throws an error if any of the dependencies are not present in the hashgraph.
 	 */
```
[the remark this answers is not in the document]
trungnotchung: updated

packages/object/src/index.ts
```diff
@@ -354,8 +371,13 @@ export class DRPObject implements ObjectPb.DRPObjectBase {
 	}
 
 	// check if the given peer has write permission
-	private _checkWriterPermission(peerId: string): boolean {
-		return this.acl ? (this.acl as ACL).query_isWriter(peerId) : true;
+	private _checkWriterPermission(peerId: string, deps: Hash[]): boolean {
+		if (!this.drp) {
+			return (this.acl as ACL).query_isWriter(peerId);
```
magnified103, line 376: What does this mean?

winprn: this handle the case where we call `_mergeWithoutDRP`

magnified103: Why can't we compute the ACL state without a DRP state?

magnified103: Resolved, unrelated to this PR

d-roak (approved):

PR #450, opened 2025-02-10
fix: serialization of map and set need to be recursive as well

sfroment:
For map and object you don't know what the values are.
So in reality if you take the example of the authorizedPeers it's a map containing Sets so it won't be serialized the way we want it if we don't make those recursive call.
Also we should go tail recursive on this one and each recursive call of this function otherwise we might encounter recursive stack limitation problem.
Will add unit test for this ASAP, but this nevertheless need to be merged,


anhnd350309 (approved):
lgtm

Issue #463, opened 2025-02-13
[FEATURE] - Avoid Redundant ACL Computation in `_initializeFinalityState`


trungnotchung:
### Problem Description

The method `_initializeFinalityState(hash: Hash)` currently recomputes the ACL even though the ACL might have already been computed and stored in the codebase. 
```typescript
// initialize the attestation store for the given vertex hash
	private _initializeFinalityState(hash: Hash) {
		if (!this.acl || !this.originalObjectACL) {
			throw new Error("ObjectACL is undefined");
		}
		const fetchedState = this.aclStates.get(hash);
		if (fetchedState !== undefined) {
			const state = cloneDeep(fetchedState);
			const acl = cloneDeep(this.originalObjectACL);

			for (const entry of state.state) {
				acl[entry.key] = entry.value;
			}
			// signer set equals writer set
			this.finalityStore.initializeState(hash, acl.query_getFinalitySigners());
		}
	}
```

`cloneDeep` operations are costly, and this function runs every time `_initializeFinalityState` is called when a new vertex is added to the hash graph. This results in unnecessary recomputation and inefficiency.

### Proposed Solution

Instead of recomputing ACL every time, modify `_initializeFinalityState` to directly accept a precomputed acl

PR #474, opened 2025-02-16
fix: redundant ACL computation

trungnotchung:
## What type of PR is this?

- [x] Optimization

## Description
* refactor `callFn` and `merge` functions that always compute `drp` and `acl` on top of both functions. So that we can utilize these pre-compute values instead of re-computing them many times.
* update `_initializeFinalityState` that can accept an acl as a parameter instead of recompute it.
* update `_computeObjectACL` to return an ACL instead of DRP.

## Related Issues and PRs

- Closes: #463 

## Added/updated tests?

- [x] No, because why: current test case covered all the cases

## Additional Info

### DIFF
--- a/packages/object/src/index.ts
+++ b/packages/object/src/index.ts
@@ -274,16 +274,18 @@ export class DRPObject implements DRPObjectBase {
 			try {
 				this.validateVertex(vertex);
 				const preComputeLca = this.computeLCA(vertex.dependencies);
-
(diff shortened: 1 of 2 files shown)


anhnd350309 (approved):
lgtm

winprn (approved):

JanLewDev (approved):
lgtm

Issue #480, opened 2025-02-18
[FEATURE] - topic discovery/heartbeat like when you are alone in the DRP topic


sfroment:
### Problem Description

When you lost conn of everyone in the topic you were in sometime you won't reconnect to any of those.

### Proposed Solution

Add an heartbeat like feature so when you are alone in the DRP so you are able to discover other peer in the topic and try to resubscribe to them

PR #502, opened 2025-02-21
feat: heartbeat initial version

sfroment:
- **feat: add message type ID heartbeat**
- **feat: initial heartbeat version**

## What type of PR is this?

- [x] Feature

## Description

When a peer is alone in an object we want it to try to find other peer subscribed to the same object

## Related Issues and PRs

- Closes: #480

## Added/updated tests?

- [x] Yes

## Additional Info


packages/node/src/heartbeat.ts
```diff
@@ -0,0 +1,224 @@
...
+		subscribers: Record<string, SubscriberInfo>
+	): Promise<void> {
+		const response = IDHeartbeatResponse.create({ subscribers });
+
+		const message = Message.create({
+			sender,
+			type: MessageType.MESSAGE_TYPE_ID_HEARTBEAT_RESPONSE,
+			data: IDHeartbeatResponse.encode(response).finish(),
+		});
+		console.log("sending message to", sender, message);
```
JanLewDev, line 221:
```suggestion
		this.logger.info(`sending message ${message} to ${sender}");
```

JanLewDev (approved):
lgtm

packages/node/src/heartbeat.ts
sfroment: :face_palm:

packages/node/src/heartbeat.ts
```diff
@@ -0,0 +1,223 @@
...
+		return true;
+	}
+
+	/**
+	 * Checks if the current search is within the configured time window.
+	 * @returns {boolean} True if within search window, false if search time has expired or not started
+	 * @private
+	 */
+	private inHeartbeatWindow(): boolean {
+		if (!this._searchStartTime) return false;
```
anhnd350309, line 129: we checked `if (!this._searchStartTime)` in pervious code, never use this code
```suggestion
```

packages/node/src/index.ts
```diff
@@ -128,13 +141,24 @@ export class DRPNode {
...
+		this.createHeartbeat(options.id);
 		return object;
 	}
 
 	async subscribeObject(id: string) {
 		await operations.subscribeObject(this, id);
 	}
 
+	private createHeartbeat(id: string): void {
+		if (this.heartbeat) this.heartbeat.stop();
```
anhnd350309, line 153: why do we need to stop heartbeat and create a new class
Why don't we stop and start again?

packages/node/src/heartbeat.ts
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Manages heartbeat functionality for distributed objects.
+ * Periodically checks and maintains connections with peers that share the same object ID.
+ * If no peers are found, it broadcasts heartbeat messages to discover peers for a configurable duration.
+ *
+ * @implements {IDRPIDHeartbeat}
+ */
+export class DRPIDHeartbeat implements IDRPIDHeartbeat {
+	private readonly _id: string;
```
anhnd350309, line 44: what _id mean in here? _id of DRP, so each DRP has DRPIDHeartbeat, right?

packages/node/src/heartbeat.ts
sfroment: it's due to lint as searchStartTime can be undef, will change the code to reflect that

packages/node/src/index.ts
sfroment: Cause the object ID might have changed

sfroment: this is related to your next comment basically, i've added some comments tell me if this is enough :)

sfroment:
Closing waiting for https://github.com/drp-tech/ts-drp/pull/524 to be merge before redoing this with this library

PR #503, opened 2025-02-21
fix: remove useless reflect call

sfroment:
Just a test


sfroment:
closing right now but need to redo it

Issue #505, opened 2025-02-23
[FEATURE] - add `packages/keychain`


trungnotchung:
### Problem Description

Currently, `DRPCredentialStore` handles private key storage. However, in cases involving a secp256k1 key, the `networkNode` also needs a matching private key. Managing these keys in multiple places can lead to inconsistencies.

### Proposed Solution

Add `packages/keychain`, and pass `secp256k1PrivateKey` as a constructor parameter in `networkNode`. Same as libp2p: https://github.com/libp2p/js-libp2p/tree/main/packages/keychain

PR #513, opened 2025-02-25
feat: add package keychain

trungnotchung:
## What type of PR is this?

- [x] Feature

## Description
* Created `packages/keychain`
* Added a new function to return `ed25519PrivateKey` in `DRPCredentialStore`. Use this private key as a parameter when starting the networkNode.
* Removed `private_key_seed` from `DRPNetworkNodeConfig`.
* Updated config files and `loadConfig` function.
* Updated relevant tests.

## Related Issues and PRs

- Closes: #505 

## Added/updated tests?

- [x] Yes

## Additional Info

### DIFF
--- a/packages/keychain/src/keychain.ts
+++ b/packages/keychain/src/keychain.ts
@@ -59,15 +59,10 @@ export class Keychain {
 		return this._blsPrivateKey.sign(uint8ArrayFromString(data)).toBytes();
 	}
 
-	get ed25519PrivateKey() {
+	get ed25519PrivateKey(): Uint8Array {
 		if (!this._ed25519PrivateKey) {
 			throw new Error("Private key not found");
 		}
-		return uint8ArrayToString(this._ed25519PrivateKey.raw, "base64");
-	}
-		if (!this._ed25519PrivateKey) {
-			throw new Error("Private key not found");
-		}
-		return uint8ArrayToString(this._ed25519PrivateKey.raw, "base64");
+		return this._ed25519PrivateKey.raw;
 	}
 }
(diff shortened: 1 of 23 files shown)


packages/keychain/src/keychain.ts
```diff
@@ -58,4 +58,11 @@ export class DRPCredentialStore {
 
 		return this._blsPrivateKey.sign(uint8ArrayFromString(data)).toBytes();
 	}
+
+	getEd25519PrivateKey() {
```
sfroment, line 62:
```suggestion
	get ed25519PrivateKey() {
		if (!this._ed25519PrivateKey) {
			throw new Error("Private key not found");
		}
		return uint8ArrayToString(this._ed25519PrivateKey.raw, "base64");
	}
```

sfroment: so like that in the code you can just call keychain1.ed25519PrivateKey

sfroment: also since privateKey is a value that only change at the construct time I think you can cache this value so we don't recalculate it over and over depending if this is often called

packages/network/src/node.ts
```diff
@@ -76,13 +75,12 @@ export class DRPNetworkNode {
...
-	async start() {
+	async start(rawPrivateKey?: string) {
 		if (this._node?.status === "started") throw new Error("Node already started");
 
 		let privateKey = undefined;
-		if (this._config?.private_key_seed) {
-			const tmp = this._config.private_key_seed.padEnd(32, "0");
-			privateKey = await generateKeyPairFromSeed("Ed25519", uint8ArrayFromString(tmp));
+		if (rawPrivateKey) {
+			privateKey = privateKeyFromRaw(uint8ArrayFromString(rawPrivateKey, "base64"));
```
sfroment, line 83: why convert this._ed25519PrivateKey.raw to string then convert back to uint8 and not passing uint8 directly ?

sfroment (changes requested):

sfroment (approved):

anhnd350309 (approved):
lgtm

PR #514, opened 2025-02-25
refactor(DRPObject): add generic type to stop using as

sfroment:

## What type of PR is this?

- [x] Refactor

## Description

Using as is highly usafe we should rely more on generic typing and what TS as too offer us

## Related Issues and PRs

- Related: #
- Closes: #

## Added/updated tests?

## Additional Info


examples/canvas/src/index.ts
JanLewDev, line 1: at drpObject def definition type is lacking 
```ts
let drpObject: DRPObject<Canvas>;
let canvasDRP: Canvas | undefined;
```

JanLewDev (approved):
otherwise lgtm

examples/canvas/src/index.ts
d-roak: this is fixed in #515

packages/object/tests/drpobject.test.ts
```diff
@@ -106,16 +106,19 @@ describe("Test for duplicate call issue", () => {
 			drp: new CounterDRP(),
 		});
 
-		const testDRP = obj.drp as CounterDRP;
+		const testDRP = obj.drp;
+		if (!testDRP) {
+			throw new Error("DRP is undefined");
+		}
```
d-roak, line 112: no need this, line 113 is checking if it's defined

packages/object/src/index.ts
```diff
@@ -34,30 +34,32 @@ export interface DRPObjectConfig {
...
-export class DRPObject implements DRPObjectBase {
+export interface DRPObjectOptions<T extends DRP> {
+	peerId: string;
+	publicCredential?: DRPPublicCredential;
+	acl?: ACL;
+	drp?: T;
+	id?: string;
+	config?: DRPObjectConfig;
+	metrics?: IMetrics;
+}
```
d-roak, line 45: this should go into `@ts-drp/types`

packages/node/src/store/object.ts
```diff
@@ -1,33 +1,37 @@
-import type { DRPObject } from "@ts-drp/object";
+/* eslint-disable @typescript-eslint/no-explicit-any -- DRPObject is not typed on purpose to allow for dynamic typing */
```
d-roak, line 1: why does it require the `T extends DRP = any` ?

packages/object/tests/hashgraph.test.ts
```diff
@@ -70,8 +70,11 @@ describe("HashGraph construction tests", () => {
 	test("Test: Vertices are consistent across data structures", () => {
 		expect(obj1.vertices).toEqual(obj1.hashGraph.getAllVertices());
 
-		const drp1 = obj1.drp as SetDRP<number>;
-		const drp2 = obj2.drp as SetDRP<number>;
+		const drp1 = obj1.drp;
+		const drp2 = obj2.drp;
+		if (!drp1 || !drp2) {
+			throw new Error("DRP is undefined");
+		}
```
d-roak, line 77: we usually shouldn't need this pattern. The fail will test further ahead, right?

packages/object/src/index.ts
```diff
@@ -77,7 +79,7 @@ export class DRPObject implements DRPObjectBase {
 				admins: new Map([[options.peerId, options.publicCredential as DRPPublicCredential]]),
 				permissionless: true,
 			});
-		this.acl = new Proxy(objAcl, this.proxyDRPHandler(DrpType.ACL));
+		this.acl = new Proxy(objAcl, this.proxyDRPHandler<ACL>(DrpType.ACL)) as ProxyHandler<ACL>;
```
anhnd350309, line 82:
```suggestion
		this.acl = new Proxy(objAcl, this.proxyDRPHandler<ACL>(DrpType.ACL)) as ACL;
```

why should we change this, and we don't need `acl as ObjectACL` anymore

anhnd350309 (approved):
otherwise lgtm

packages/object/tests/hashgraph.test.ts
sfroment: yes I can remove this and only use ?

packages/node/src/store/object.ts
sfroment: any here is necessary to it can be used in DRPNode in a nicer way.
But reflecting on that I think a better way would be to make get/put generic and have store like `Map<string, DRPObject<any>>`

trungnotchung (approved):
lgtm

winprn (approved):

packages/object/src/index.ts
sfroment: done

sfroment:
closing this one will reopen one with evertything updated from scratch

PR #515, opened 2025-02-25
refactor: add type on all function

sfroment:

## What type of PR is this?

- [x] Refactor

## Related Issues and PRs

- Related: #
- Closes: #

## Added/updated tests?

## Additional Info

### DIFF
--- a/packages/network/src/stream.ts
+++ b/packages/network/src/stream.ts
@@ -23,12 +23,12 @@ import * as lp from "it-length-prefixed";
 import map from "it-map";
 import { pipe } from "it-pipe";
 
-export async function uint8ArrayToStream(stream: Stream, input: Uint8Array) {
+export async function uint8ArrayToStream(stream: Stream, input: Uint8Array): Promise<void> {
 	await pipe(input, (source) => lp.encode([source]), stream.sink);
 }
 
-export async function streamToUint8Array(stream: Stream) {
-	return await pipe(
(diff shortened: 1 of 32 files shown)


trungnotchung (approved):
lgtm

sfroment:
will merge this one once the the DRPObject Generic is reviewed

packages/network/src/stream.ts
```diff
@@ -23,12 +23,12 @@ import * as lp from "it-length-prefixed";
...
 
-export async function uint8ArrayToStream(stream: Stream, input: Uint8Array) {
+export async function uint8ArrayToStream(stream: Stream, input: Uint8Array): Promise<void> {
 	await pipe(input, (source) => lp.encode([source]), stream.sink);
 }
 
-export async function streamToUint8Array(stream: Stream) {
-	return await pipe(
+export async function streamToUint8Array(stream: Stream): Promise<Uint8Array> {
+	return pipe(
```
d-roak, line 31: why are we taking the await on one and not the other?

sfroment: I can change the rule to always, but basically the rules `"@typescript-eslint/return-await": ["error", "in-try-catch"]` only force to await a promise in a try catch cause if you try catch and don't await you loose the control of the promise you return. There as it's not in a try catch there is no need to await therefore it only add complexity to the code.

PR #519, opened 2025-02-26
fix: linting deps not at the correct place

sfroment:

## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related Issues and PRs

- Related: #
- Closes: #

## Added/updated tests?

- [ ] Yes
- [ ] I need help with writing tests

## Additional Info


.prettierrc.json
```diff
@@ -8,5 +8,14 @@
 	"trailingComma": "es5",
 	"bracketSpacing": true,
 	"arrowParens": "always",
-	"endOfLine": "auto"
+	"endOfLine": "auto",
+	"overrides": [
+		{
+			"files": "package.json",
+			"options": {
+				"useTabs": false,
```
d-roak, line 16: it shouldn't be tabs true?

sfroment: it is more common in package.json to use space and this is why we have a problem with https://github.com/release-it-plugins/workspaces writing after ward. Detect indent only detect the width of the indent so if we use tabs length will be one. And so this is why on write https://github.com/release-it-plugins/workspaces/blob/master/index.js#L133 it will override tabs to space with a width of 1. 
Nevertheless we can still useTabs and just prettier the stuff after a release in the pnpm lint:fix, LMK what you want :)

trungnotchung (approved):
lgtm

anhnd350309 (approved):
lgtm

PR #524, opened 2025-02-28
feat: add interval-runner

sfroment:
- **feat: add package utils**
- **feat: add package interval-runner**


JanLewDev (approved):
lgtm

packages/interval-runner/src/index.ts
```diff
@@ -0,0 +1,113 @@
...
+	 */
+	start(args?: Args): void {
+		if (this._state === 1) {
+			throw new Error("Interval runner is already running");
+		}
+
+		this._state = 1;
+
+		const scheduleNext = async () => {
+			if (this._state === 0) return;
```
anhnd350309, line 72: this code is never run

packages/interval-runner/tests/interval-runner.test.ts
anhnd350309, line 1: we need to add some tests
- should stop when function throws error
- should have arg in function
example
```
describe("should stop when function throws error", () => {
	it("should stop when function throws error", async () => {
		const callback = vi.fn().mockImplementation(() => {
			throw new Error("Test error");
		});
		const runner = new IntervalRunner({ interval: 1000, fn: callback });

		runner.start();
		expect(callback).toHaveBeenCalledTimes(1); // Should be called immediately
		expect(runner.state).toBe("running");

		await vi.advanceTimersByTimeAsync(1000);
		expect(callback).toHaveBeenCalledTimes(1); // Should not be called again
		expect(runner.state).toBe("stopped");
	});
});

describe("should have arg in function", () => {
	it("should pass arg to function", async () => {
		const callback = vi.fn().mockImplementation((arg: number) => arg);
		const runner = new IntervalRunner({ interval: 1000, fn: callback });
		// eslint-disable-next-line @typescript-eslint/no-explicit-any
		const args: any = [10];
		runner.start(args);
		expect(callback).toHaveBeenCalledTimes(1); // Should be called immediately
		expect(runner.state).toBe("running");
	});
});
```

packages/interval-runner/package.json
```diff
@@ -0,0 +1,37 @@
...
+		"cli": "tsx ./src/run.ts",
+		"prebuild": "node -p \"'export const VERSION = ' + JSON.stringify(require('./package.json').version) + ';'\" > src/version.ts",
+		"prepack": "tsc -b",
+		"test": "vitest",
+		"watch": "tsc -b -w"
+	},
+	"dependencies": {
+		"@ts-drp/logger": "0.8.5",
+		"@ts-drp/types": "0.8.5",
+		"@ts-drp/utils": "workspace:^0.8.5"
```
trungnotchung, line 35:
```suggestion
		"@ts-drp/utils": "0.8.5"
```

packages/interval-runner/src/index.ts
```diff
@@ -0,0 +1,125 @@
...
+			crypto
+				.createHash("sha256")
+				.update(Math.floor(Math.random() * Number.MAX_VALUE).toString())
+				.digest("hex");
+	}
+
+	private async execute(args?: Args): Promise<boolean> {
+		const result = args ? this.fn(...args) : this.fn();
+
+		if (isAsyncGenerator(result)) {
```
trungnotchung, line 44: why are we making decisions based on only the last value instead of all?

trungnotchung: gotcha

trungnotchung (approved):
lgtm

anhnd350309 (approved):
lgtm

PR #528, opened 2025-03-04
feat: interval-drp-discovery

sfroment:
## What type of PR is this?

- [x] Feature

## Description

Add interval drp discovery, replace the #502 

## Related Issues and PRs

- Related: #502
- Closes: #480

## Added/updated tests?

- [x] Yes

## Additional Info


sfroment:
Will be marked as ready when interval-runner is merged

packages/types/src/interval-discovery.ts
```diff
@@ -0,0 +1,28 @@
+export interface DRPDiscovery {
```
anhnd350309, line 1: Do we need this interface

packages/node/src/index.ts
```diff
@@ -155,4 +181,35 @@ export class DRPNode {
...
+				...this.config.interval_discovery_options,
+				id,
+				networkNode: this.networkNode,
+			});
+
+		this._intervals.set(id, interval);
+		interval.start();
+	}
+
+	async handleDiscoveryResponse(sender: string, data: Uint8Array): Promise<void> {
```
trungnotchung, line 201: why do we handle this case inside the node instead of the handler?

packages/interval-discovery/src/index.ts
```diff
@@ -0,0 +1,260 @@
...
+		const defaultSearchDuration = 5 * 60 * 1000; // 5 minutes
+		this.networkNode = opts.networkNode;
+		this.searchDuration = opts.searchDuration ?? defaultSearchDuration;
+		this._logger = new Logger(`drp::discovery::${opts.id}`, opts.logConfig);
+		// Create the delegate interval runner
+		this._intervalRunner = new IntervalRunner({
+			...opts,
+			fn: async (): Promise<boolean> => {
+				await this._runDRPDiscovery();
+				return true; // Always continue the interval
```
trungnotchung, line 50: why this function always return true to continue the `_intervalRunner`. i think make _runDRPDiscovery to return a boolean is better.

sfroment: I was just not willing to add ret true everywhere in the run, updated

packages/interval-discovery/src/index.ts
```diff
@@ -0,0 +1,260 @@
...
+	private _intervalRunner: IntervalRunner;
+
+	/** Logger instance with discovery-specific prefix */
+	private _logger: Logger;
+
+	/**
+	 * Creates a new DRP Discovery instance
+	 */
+	constructor(opts: DRPIntervalDiscoveryOptions) {
+		const defaultSearchDuration = 5 * 60 * 1000; // 5 minutes
```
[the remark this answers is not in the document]
sfroment: updated

packages/types/src/interval-discovery.ts
sfroment: oups forgot to delete this one

anhnd350309 (approved):

packages/interval-discovery/src/index.ts
```diff
@@ -0,0 +1,258 @@
...
+
+	/**
+	 * Runs a single discovery cycle to find and connect with peers
+	 */
+	private async _runDRPDiscovery(): Promise<boolean> {
+		// Early exit if we already have peers
+		if (this._hasPeers()) {
+			this._searchStartTime = undefined;
+			return true;
+		}
```
trungnotchung, line 64: doesn't we should return false here?

sfroment: Cause we still want to periodically check that you indeed have peer, but it won't fire any request what so ever if you have peer. It will just reschedule a check for this to be recalled again

trungnotchung (approved):
lgtm

packages/interval-discovery/src/index.ts
```diff
@@ -0,0 +1,258 @@
...
+	Message,
+	MessageType,
+	type SubscriberInfo,
+} from "@ts-drp/types";
+
+const DEFAULT_SEARCH_DURATION = 5 * 60 * 1000; // 5 minutes
+/**
+ * Enhanced DRP Discovery service using composition pattern
+ * Implements IntervalRunnerInterface to maintain compatibility with IntervalRunner[] arrays
+ */
```
JanLewDev, line 19: This comment should be in types
```ts
export interface IDRPIntervalDiscovery extends IIntervalRunner<"interval:discovery"> {
```

JanLewDev (approved):
lgtm

Issue #479, opened 2025-02-18
[FEATURE] - Remove public keys from ACL grant operations


magnified103:
### Problem Description

Currently, the ACL grant operation (`grant(peerId, pubkey)`) requires the BLS public key as a parameter.

It is possible to remove the pubkey parameter from the grant operation, as the peers should set it themselves.

### Proposed Solution

Change the grant operation to `grant(peerId, pubkey)`.

Add a new ACL operation named `setKey(pubkey)`. This operation sets the BLS pubkey for its owner.

PR #536, opened 2025-03-05
feat: update object acl 

trungnotchung:
## What type of PR is this?

- [x] Feature

## Description
- Removed publicKey from the constructor &  `grant` function.
- Adddd function `setKey` for node to set their publicKey.
- Updated `DRPObject` to only verify writer permission when `operation.drpType` is DRP
- Updated relevant tests

## Related Issues and PRs

- Closes: #479 

## Added/updated tests?

- [x] Yes


packages/object/src/acl/index.ts
```diff
@@ -19,7 +34,7 @@ export class ObjectACL implements IACL {
 	private _authorizedPeers: Map<string, PeerPermissions>;
 
 	constructor(options: {
-		admins: Map<string, DRPPublicCredential>;
+		admins: string[];
```
anhnd350309, line 37: Why do we change this? We can init keychain in here and still have the `setKey` method

trungnotchung: it's unnecessary. remove it can make things simpler

trungnotchung: also with the setKey function, only node can setKey for itself

anhnd350309 (approved):
lgtm

sfroment (approved):

Issue #511, opened 2025-02-24
[BUG] - fix missing timeout in `handlers.test.ts`


trungnotchung:
### Description

Currently, tests in `handler.test.ts` are not working as intended. The message we send in the test does not cause changes to the nodes. When the drp is updated, `drpObjectChangesHandler` will be called and will broadcast messages to the network. We just need to timeout right after the update to see the changes.

### Version

0.7.0

### Reproduction steps

```bash
1. Go to `handler.test.ts`
2. Add this code right after updating the drp:

await new Promise((resolve) => setTimeout(resolve, 5000));
```

### Screenshots

![DESCRIPTION](LINK.png)

### Logs

```bash

```

sfroment:
this on

PR #557, opened 2025-03-12
Fix: testcase incorrect

anhnd350309:
## What type of PR is this?

Make testcase correct
- Make unit test independence
- Remove send message unuse (replace timeout to wait message send)

- [x] Refactor

## Related Issues and PRs

- Related: #511 
- Closes: #

## Added/updated tests?

- [ ] Yes: but only update testcase

## Additional Info

### DIFF
--- a/packages/node/tests/handlers.test.ts
+++ b/packages/node/tests/handlers.test.ts
@@ -2,7 +2,7 @@ import type { Connection, IdentifyResult, Libp2p, Stream } from "@libp2p/interfa
 import { SetDRP } from "@ts-drp/blueprints";
 import { DRPNetworkNode, type DRPNetworkNodeConfig } from "@ts-drp/network";
 import { type DRPObject, ObjectACL } from "@ts-drp/object";
-import { DrpType, FetchState, Message, MessageType, Sync, Update } from "@ts-drp/types";
+import { DrpType, FetchState, Message, MessageType, Sync } from "@ts-drp/types";
 import { raceEvent } from "race-event";
 import { afterAll, beforeAll, beforeEach, describe, expect, test, vi } from "vitest";
 
--- a/packages/node/tests/handlers.test.ts
+++ b/packages/node/tests/handlers.test.ts
@@ -147,12 +147,11 @@ describe("Handle message correctly", () => {
 			id: drpObjectNode2.id,
 			acl: acl,
 		});
-
-		(drpObjectNode2.drp as SetDRP<number>).add(5);
-		(drpObjectNode2.drp as SetDRP<number>).add(10);
 	});
 
 	test("should handle update message correctly", async () => {
+		(drpObjectNode2.drp as SetDRP<number>).add(5);
+		(drpObjectNode2.drp as SetDRP<number>).add(10);
 		const vertices = drpObjectNode2.vertices;
 		await signGeneratedVertices(node2, vertices);
 		const message = Message.create({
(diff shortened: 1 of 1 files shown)


packages/node/tests/handlers.test.ts
```diff
@@ -158,34 +146,36 @@ describe("Handle message correctly", () => {
...
 		await node1.createObject({
 			drp: new SetDRP<number>(),
-			id: drpObject.id,
+			id: drpObjectNode2.id,
 			acl: acl,
 		});
 
-		(drpObject.drp as SetDRP<number>).add(5);
-		(drpObject.drp as SetDRP<number>).add(10);
+		(drpObjectNode2.drp as SetDRP<number>).add(5);
```
sfroment, line 159: I don't think that this shall be in the before each with that you do ops in the before each which is weird

sfroment (changes requested):

packages/node/tests/handlers.test.ts
anhnd350309: If I don't set in the before each, I need to add operation in each testcase. but I think I need to change it

anhnd350309: updated

packages/node/tests/handlers.test.ts
```diff
@@ -142,42 +133,40 @@ describe("Handle message correctly", () => {
...
 			type: MessageType.MESSAGE_TYPE_UPDATE,
 			data: Update.encode(
 				Update.create({
-					objectId: drpObject.id,
+					objectId: drpObjectNode2.id,
 					vertices: vertices,
 				})
 			).finish(),
 		});
 		await node2.networkNode.sendMessage(node1.networkNode.peerId, message);
```
trungnotchung, line 167: these code doesnt do anything.

packages/node/tests/handlers.test.ts
```diff
@@ -188,42 +177,48 @@ describe("Handle message correctly", () => {
...
 			sender: node1.networkNode.peerId,
 			type: MessageType.MESSAGE_TYPE_SYNC,
 			data: Sync.encode(
 				Sync.create({
-					objectId: drpObject.id,
-					vertexHashes: node1.objectStore.get(drpObject.id)?.vertices.map((vertex) => vertex.hash),
+					objectId: drpObjectNode2.id,
+					vertexHashes: node1.objectStore
+						.get(drpObjectNode2.id)
+						?.vertices.map((vertex) => vertex.hash),
```
trungnotchung, line 221: these code doesnt do anything.

trungnotchung (commented):
we only have two nodes that already connected to each other and they will update right after drp in one node is updated. i think its best to create a new node to connect with these node and send message to test.

packages/node/tests/handlers.test.ts
anhnd350309: sync message cannot sent automatically, so we still need send message Sync.

packages/node/tests/handlers.test.ts
anhnd350309: yup, updated

anhnd350309:
> we only have two nodes that already connected to each other and they will update right after drp in one node is updated. i think its best to create a new node to connect with these node and send message to test.
In this PR, I just wanna fix this independent testcase, the next PR, I'll add handle event and create other node to accept message

packages/node/tests/handlers.test.ts
trungnotchung: yes it does not send automatically. but sending this message in this test does not change anything

anhnd350309: no, It'll change the state of the node, I have expected a test case after this line

trungnotchung: the state changed not because of the message. you just forgot to await

anhnd350309: updated

trungnotchung (approved):
lgtm

sfroment (approved):

Issue #121, opened 2024-08-26
Add message queues


d-roak:
In the future, we might reach a bottleneck on how many messages we can process considering the volume of incoming messages. We want to implement message queuing to address this

PR #566, opened 2025-03-14
feat!: add message queues

JanLewDev:
## What type of PR is this?

- [x] Feature

## Description

Introduces message queues.
- move `object_id` to the general message in protobuf

## Related Issues and PRs

- Related: #
- Closes: #121 

## Added/updated tests?

- [x] Yes


packages/message-queue/src/message-queue.ts
```diff
@@ -0,0 +1,77 @@
...
+				}
+			} catch (error) {
+				// When the channel is closed, exit the loop.
+				if (error instanceof Error && error.message === "Channel is closed") {
+					break;
+				} else {
+					console.error("Error in fanout loop:", error);
+				}
+			}
+		}
```
JanLewDev, line 67:
```suggestion
        if (this.fanoutLoopStarted) return;
        this.fanoutLoopStarted = true;
		while (this.isActive) {
			try {
				const message = await this.channel.receive();

				for (const handler of this.subscribers) {
					try {
						await handler(message);
						console.log(`queue::processed message ${message}`);
					} catch (error) {
						console.error(`Error in handler processing message ${message}:`, error);
					}
				}
			} catch (error) {
				// When the channel is closed, exit the loop.
				if (error instanceof Error && error.message === "Channel is closed") {
					break;
				} else {
					console.error("Error in fanout loop:", error);
				}
			}
		}
        this.fanoutLoopStarted = false;
```

packages/types/src/message-queue.ts
```diff
@@ -0,0 +1,63 @@
+import type { LoggerOptions } from "@ts-drp/logger";
```
JanLewDev, line 1:
```suggestion
import type { LoggerOptions } from "./logger.js";
```

Issue #378, opened 2025-01-20
Add `packages/validation`


d-roak:
Centralize the validation rules
Extra points: implement some type of decorator for object validation

Issue #488, opened 2025-02-19
[BUG] - Message handler error when DRP id is an empty string


winprn:
### Description

When we join a DRP with empty id, the messages became wild and it keeps throwing errors. This should be validate to not allow creating empty id DRP

### Version

latest

### Reproduction steps

```bash
1. Create an empty id grid DRP, by clicking the JOIN button
2. Start another node and click JOIN again
3. View the error in console
```

### Logs

```bash

```

hoangquocvietuet:
We will drop the DRP id afaik so it won't happen

d-roak:
add this checks on #378

Issue #377, opened 2025-01-20
Add `packages/errors`


d-roak:
Add custom error handlers for the whole codebase with error codes and messages
Extra points: initial i18n implementation

trungnotchung:
move `packages/validation/error.ts`to this package

PR #617, opened 2025-03-25
feat: add package validation

trungnotchung:
## What type of PR is this?

- [x] Feature

## Description
- Validated incoming message using `zod`
- Validated `createObject` and `connectObject` options using `zod`
- Moved `validateVertex` to packages/validation
- Updated relevant tests and added tests.

## Related Issues and PRs

- Closes: #378 , #488 

## Added/updated tests?

- [x] Yes

## Additional Info


packages/object/tests/hashgraph.test.ts
```diff
@@ -173,19 +174,21 @@ describe("HashGraph construction tests", () => {
 			Date.now(),
 			new Uint8Array()
 		);
-		expect(() => {
-			obj1.validateVertex(fakeRoot);
```
JanLewDev, line 177: these should ba in the `validation.test.ts`

packages/object/src/index.ts
```diff
@@ -349,9 +319,23 @@ export class DRPObject<T extends IDRP> implements DRPObjectBase, IDRPObject<T> {
 			}
 
 			try {
-				this.validateVertex(vertex);
+				const validation = validateVertex(vertex, this.hashGraph, Date.now());
+				if (!validation.success) {
+					throw new Error(validation.error);
```
sfroment, line 324: what is the real type of validation.error before you take the error.Message ? Maybe there is no point of creating an error again from a string extracted from an error ?

packages/object/src/index.ts
```diff
@@ -349,9 +319,23 @@ export class DRPObject<T extends IDRP> implements DRPObjectBase, IDRPObject<T> {
...
+				}
 				const preComputeLca = this.computeLCA(vertex.dependencies);
 
+				const acl = this._computeObjectACL(
+					vertex.dependencies,
+					preComputeLca,
+					vertex.operation.drpType === DrpType.ACL ? vertex.operation : undefined,
+					vertex.peerId
+				);
+				if (vertex.operation?.drpType === DrpType.DRP && !acl.query_isWriter(vertex.peerId)) {
```
sfroment, line 334: not really related to this PR, also shouldn't we check ACL even before ?

packages/node/src/index.ts
```diff
@@ -122,6 +125,14 @@ export class DRPNode {
 	}
 
 	async createObject<T extends IDRP>(options: NodeCreateObjectOptions<T>): Promise<DRPObject<T>> {
+		if (this.networkNode.peerId === "") {
+			throw new Error("Node not started");
+		}
+		const validation = NodeCreateObjectOptionsSchema.safeParse(options);
+		if (!validation.success) {
+			throw new Error(`Invalid options when creating object: ${validation.error.message}`);
```
sfroment, line 133: The zod error as a lot of usefull info I don't like extracting only a part of it. If you which to still have a small message in before you could get a small helper class like 
```ts
class DRPValidationError extends Error {
    zodError: ZodError;

    constructor(zodError: ZodError) {
        super(zodError.message);
        this.zodError = zodError;
        this.name = "DRPValidationError";
    }
}

```

sfroment (changes requested):

packages/object/src/index.ts
trungnotchung: this condition is from the old function `validateVertex`. i check it after computing the `acl` because this function also needs to calculate the acl. so doing this can avoid redundancy computation

packages/object/tests/hashgraph.test.ts
trungnotchung: updated

packages/node/src/index.ts
trungnotchung: updated

packages/object/src/index.ts
trungnotchung: updated

sfroment (approved):

trungnotchung:
we should move the errors to #377

JanLewDev (approved):
lgtm

PR #618, opened 2025-03-25
refactor: move object/utils to package utils

trungnotchung:
## What type of PR is this?

- [x] Refactor

## Description
Reason: better code structure, avoid circular dependency
- Moved folder utils in package object to package utils
- Updated relevant import

## Added/updated tests?

- [x] No, because it's just a refactor

## Additional Info


JanLewDev (approved):

packages/utils/src/index.ts
```diff
@@ -172,3 +172,7 @@ export function handlePromiseOrValue<T, R>(value: T | Promise<T>, fn: (value: T)
 	}
 	return fn(value);
 }
+
+export * from "./object/computeHash.js";
```
sfroment, line 176: We don't want that cause this mean that when you import only one for the function of index you will have all those code imported as well (meaning that the deps msgpack and hashes will be include in it). Could you instead change the package.json export and make thing seperate. 
Ideally we want one for the serializer as this import msgpack, one for the compute hash as this one import noble hashes. And ofc a last one for objectSet even though for object set as this one as no deps it might be okay to keep it here

sfroment (changes requested):
see comment above

packages/utils/src/index.ts
trungnotchung: updated

packages/node/tests/utils.test.ts
```diff
@@ -1,5 +1,6 @@
 import { SetDRP } from "@ts-drp/blueprints";
-import { deserializeDRPState, DRPObject, serializeDRPState } from "@ts-drp/object";
+import { DRPObject } from "@ts-drp/object";
+import { deserializeDRPState, serializeDRPState } from "@ts-drp/utils/serialization";
```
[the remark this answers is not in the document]
trungnotchung: kiwi's comment explained that :v

anhnd350309 (approved):
lgtm

sfroment (approved):

Issue #622, opened 2025-03-26
[FEATURE] - Chat example refactor


sfroment:
### Problem Description

The goal of this issue is to refactor the chat examples.
Right now the styling was done with AI

### Proposed Solution

what we want:
- better styling
- peer id shortened and colored
- mobile friendly
- display error with toast message so people can see the error in a good way
- maybe use the unpkg ? So we only have an index.html ? (Maybe)

Anything that will make this more pretty

PR #631, opened 2025-03-30
feat(examples): chat example revamp 

JanLewDev:
## What type of PR is this?

- [x] Feature

## Description

Features:
- [x] toast messages
- [x] copy id's to clipboard
- [x] better mobile support
- [x] dark mode 

## Related Issues and PRs

- Related: #622 

## Added/updated tests?

- [x] No, because style changes

## Screenshots

obraz
obraz
obraz
obraz

### DIFF
--- a/examples/chat/index.html
+++ b/examples/chat/index.html
@@ -9,6 +9,10 @@
 	<body>
 		<h1>DRP - Chat</h1>
 		
+		<button id="themeToggle" class="theme-toggle" aria-label="Toggle dark mode">
+			🌙
+		</button>
+		
 		<div class="info-row">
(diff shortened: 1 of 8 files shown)


sfroment (changes requested):
Really nice, although few things to fix:
- when clicking the create button and that I've put something in the input it will create the room with a random id
	- for that we would need to either put the button on top of the input and write something like create random room ? 
- when cliking on the copy button of a room it will paste like so `518c...20ea` so not the full room id
- flickering at start when I'm already in dark (I've pushed a fix for that)

sfroment (approved):

anhnd350309 (approved):

PR #644, opened 2025-04-02
fix: try to have consistency in test

sfroment:
- **fix: we only want one ope and wait for an attestation update on the vertices of this test**
- **chore: increase a bit the timeout of flaky test**

## What type of PR is this?

- [x] Bug Fix

## Description

I've might have been a bit to hard reducing timeout from 60s to 10s let's try increasing this a bit and see how it goes.
Additionaly one other thing is the fix on the latest test we only want to add one thing and then wait for the attestation

## Related Issues and PRs

- Related: #
- Closes: #

## Added/updated tests?

## Additional Info


anhnd350309 (approved):
lgtm

PR #647, opened 2025-04-04
fix(examples): handle wsad correctly in grid

JanLewDev:
## What type of PR is this?

- [x] Bug Fix

## Description

Currently even if grid not initialised or when initialised but in an input field such errors arise in the console.
obraz

## Related Issues and PRs

- Related: #
- Closes: #

## Added/updated tests?

- [x] No, because not needed

## Additional Info

### DIFF
--- a/examples/grid/src/state.ts
+++ b/examples/grid/src/state.ts
@@ -28,7 +28,6 @@ class GridStateManager implements GridState {
 
 	isNodeInitialized(): boolean {
 		if (!this.node) {
-			console.error("Node not initialized");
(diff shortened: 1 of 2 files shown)


examples/grid/src/state.ts
```diff
@@ -28,15 +28,13 @@ class GridStateManager implements GridState {
 
 	isNodeInitialized(): boolean {
 		if (!this.node) {
-			console.error("Node not initialized");
```
sfroment, line 31: warn but keep them

